PR #25, opened 2021-04-14
fix: add break-word so selected option does not get cut off

snapsnapturtle:
Copy from the internal repository


lukahartwig:
@all-contributors please add @snapsnapturtle for code doc bug review

Issue #27, opened 2021-04-16
PhoneInput test is running slow


lukahartwig:
One of the PhoneInput tests is running very slow for some reason. This slows down the build and also leads to timeout issues in the CI pipeline.

<img width="623" alt="Screenshot 2021-04-16 at 10 41 09" src="https://user-images.githubusercontent.com/7414521/114997714-6fa56b80-9ea0-11eb-952a-12199f04dd8c.png">

PR #30, opened 2021-04-16
build: disabled the slow running tests that made the pipeline fail

lukahartwig:
This PR disabled the slow running test that makes the pipeline fail.

I created an issue to track this test here #27


lukahartwig:
@all-contributors please add @lukahartwig for code doc bug review

PR #34, opened 2021-04-21
docs: removed some pages and internal references

lukahartwig:
Removed SECURITY and MAINTAINERS page which were added
accidentally, as well as a link to intenal slack.


duranmla (approved):
👍

PR #40, opened 2021-04-23
Refactor/adapt selectlist style

div-Leo:
**What**
This PR aims to adapt the `SelectList` component style to the given [design](https://app.zeplin.io/project/57fe438f81b870887cc71736/screen/605c7840aebc4544301abe77)

**How**
Update the color of the selected items and add the style for the disabled options.
Update the example showing the disabled options.

* feat: add disabled option style
* refactor: apply updated design
Screenshot 2021-04-20 at 16 15 01
Screenshot 2021-04-20 at 16 13 57

### DIFF
--- a/src/components/SelectList/SelectList.tsx
+++ b/src/components/SelectList/SelectList.tsx
@@ -1,9 +1,8 @@
-import React, { FC, useEffect, useRef, useState } from 'react';
+import React, { FC } from 'react';
 import ReactSelect, { components as ReactSelectComponents, IndicatorProps, Props, StylesConfig } from 'react-select';
-import { MarginProps, style, WidthProps } from 'styled-system';
+import { MarginProps, WidthProps } from 'styled-system';
 
 import { Colors, Elevation } from '../../essentials';
-import { theme } from '../../essentials/theme';
 import { ChevronDownIcon, ChevronUpIcon, CloseIcon } from '../../icons';
 import {
     ClassNameProps,
--- a/src/components/SelectList/SelectList.tsx
+++ b/src/components/SelectList/SelectList.tsx
@@ -1,6 +1,6 @@
 import React, { FC, useEffect, useRef, useState } from 'react';
 import ReactSelect, { components as ReactSelectComponents, IndicatorProps, Props, StylesConfig } from 'react-select';
-import { MarginProps, WidthProps } from 'styled-system';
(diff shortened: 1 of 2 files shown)


src/components/SelectList/SelectList.tsx
```diff
@@ -131,21 +130,41 @@ const customStyles: StylesConfig = {
...
+            .filter(key => state[key])
+            .reduce((acc, style) => ({ ...acc, ...colorsByState[style] }), defaultColors);
+
+        return {
+            ...provided,
+            ...colors,
+            '&:active': {
+                ...colors
+            },
+            padding: `${state.selectProps.size === 'medium' ? '0.5rem' : '0.375rem'} 0.75rem`,
```
duranmla, line 162: While I understand that spaces should be only used on margins I wonder if paddings should have a certain guideline, padding of `0.375rem` looks arbitrary. @agonzalezUX any thoughts on this?

duranmla (changes requested):
I would like to align with Designers to make sure about this decision on padding sizes.
Thanks a lot for coming to contribute @div-Leo

src/components/SelectList/SelectList.tsx
agonzalezUX: we don't have paddings set in our library but they should follow the same size and structure as the margins.

agonzalezUX: here we should use or spacing (padding) 1 (0.5rem) or half of it which would be 0.25rem

duranmla: @agonzalezUX that means we need to make sure [these types of designs](https://app.zeplin.io/project/57fe438f81b870887cc71736/screen/605c7840aebc4544301abe77) get updated.

For small will be `0.25rem`, medium `0.5rem` and large `1rem`

agonzalezUX: ohh wait, my bad. Please forget what I said. I think that before use the same scale as the spacing we need to review that this is matching with the components that has padding so, for now I would not change it and would keep using what is in the zeplin file. Sorry, my fault 🙏🏼

duranmla (approved):
After @agonzalezUX comment then I believe we are good to go with this one.

duranmla:
@all-contributors please add @div-Leo for his contribution on code Thanks a lot mate!

PR #41, opened 2021-04-23
Feat/add checkbox variants

div-Leo:
**What** 
This PR aims to add two size variants for the checkbox.

**Why**
The design was providing two more sizes that the library was missing.

**How**
Add a `size` prop to the component with variants "small" | "medium" | "large" and pass it to the `Text` component used as the label.
Defaults to 1rem ("medium")

Screenshot 2021-04-21 at 16 47 45

* feat: add small and medium size variants
* test: add snapshots test for the 3 size variants


duranmla:
@div-Leo Thanks for coming here! This will allow you to spot some details we need to adjust to allow people to easily contribute. With that being said:

1. Can you take a look at your screenshot link? (it seems to be pointing to something other than the image)
2. _The changes look good, I will try to take a look to understand what's going on with the build_

agonzalezUX:
hi! One question, is this checkbox aligned left-top? or is still vertically centered when there is a text block?

div-Leo:
I'm not 100% sure I got the question about the alignment, but without the text label, it will still be top-left.

Here you have couple of media as reference, maybe it clarifys the doubt.
Screenshot 2021-04-23 at 15 04 50

https://user-images.githubusercontent.com/12762609/115875367-6dfe1980-a445-11eb-8b39-32a720f2bd12.mov

agonzalezUX:
> I'm not 100% sure I got the question about the alignment, but without the text label, it will still be top-left.
> 
> Here you have couple of media as reference, maybe it clarifys the doubt.
> Screenshot 2021-04-23 at 15 04 50
> 
>  Screen.Recording.2021-04-23.at.15.01.35.mov

I mean, when there is a block of text, more than two three lines like we have in the terms and conditions for example, is the checkbox aligned top left still?

thanks! :)

div-Leo:
We have two versions based on a property we can pass to the component called `textVerticalAlign`. By default is `"center"` which means that is going to be centered with the text (fig. 1). Otherwise, we can set it to be `"top"` which aligns it to the top (fig. 2)

Screenshot 2021-04-26 at 10 11 41
Screenshot 2021-04-26 at 10 11 29

Hope this clarifies 😊

agonzalezUX:
> We have two versions based on a property we can pass to the component called `textVerticalAlign`. By default is `"center"` which means that is going to be centered with the text (fig. 1). Otherwise, we can set it to be `"top"` which aligns it to the top (fig. 2)
> 
> Screenshot 2021-04-26 at 10 11 41
> 
> Screenshot 2021-04-26 at 10 11 29
> 
> Hope this clarifies 😊

niceee!!! I think tho that the default one should be the top aligned ^_^

duranmla:
All right! It seems we are good to go with this. We can update with the main branch using the UI and merge.

duranmla (approved):

lukahartwig:
One thing to consider is that `label` also can take arbitrary react elements as input. This might break the behavior of `textVerticalAlign` when we change the default to `top`. I would changing the default value therefore a breaking change and would like to avoid this for now.

Screenshot 2021-04-26 at 14 07 54

src/components/Checkbox/Checkbox.tsx
```diff
@@ -65,4 +69,10 @@ const Checkbox: FC<CheckboxProps> = props => {
     );
 };
 
+Checkbox.defaultProps = {
+    textVerticalAlign: 'top',
```
lukahartwig, line 73: See my comment on the PR why I consider this a breaking change. @duranmla do you agree with this?

lukahartwig (changes requested):
Just a clarification on the change of the default behavior.

lukahartwig:
@div-Leo would you mind also adding an entry to the changelog, please?

src/components/Checkbox/Checkbox.tsx
duranmla: Got it, I understand your point. But it almost feels like we need to comment on the "breaking" change on the documentation instead, isn't it? As you are adding a custom element we can't guarantee you won't break things. Maybe we can keep the default as `top` because also the default user is not using custom labels.

Besides, I think we should strive for making pieces that own their space without affecting the exterior. A headline maybe should not have an inherent margin?

duranmla: My point above is related to https://youtu.be/xxbc3wAztl0?t=551 _" components shouldn't contain surrounding white space"_ in that sense I wouldn't shape a component to meet the opposite of the quote I just give to you here.

lukahartwig: I agree that the default behavior SHOULD be `top`, but supporting custom elements was a use case we explicitly wanted and do support currently. Headline is just an example but more common is to use `Checkbox` with an icon and text combination for example.

The `top` option only works in very limited situation so with the current implementation of `Checkbox` I don't think we should make this default.

I definitely think this is something we should fix in the next major version though.

> Besides, I think we should strive for making pieces that own their space without affecting the exterior. A headline maybe should not have an inherent margin?

The problem is not the margin. The checkbox should be centered on the first line of the text but sticks to the top. So when the checkbox is significantly smaller or bigger than line-height he label, it will look broken.

agonzalezUX:
on "One thing to consider is that label also can take arbitrary react elements as input." I understand the flexibility of this component but, we should only use it as specified on the style guide. So in this case we should avoid using other text-styles than the one specified :)

lukahartwig:
> on "One thing to consider is that label also can take arbitrary react elements as input." I understand the flexibility of this component but, we should only use it as specified on the style guide. So in this case we should avoid using other text-styles than the ones specified :)

I agree, but we DO use it currently this way which makes it hard to change without breaking existing behavior.

We could also bite the bullet and make a breaking change with the release of the open source version. But in that case we should have a migration plan for the internal users.

src/components/Checkbox/Checkbox.tsx
div-Leo: So to avoid breaking changes shall we revert to have the default to be `center`?

div-Leo:
About breaking changes I also left a note (TODO) about the default size. 
By design, it should be `1rem`, considered to be the `large` variant, but in the current version `large` corresponds to `1.125rem` and `medium` to `1rem`. 
To not create differences in the usage of large I followed the same behavior we have now with the `Text` component and I set the default to be `medium` (i.e. `1rem`) which is the current size of the checkbox labels (before the PR).

I'm going to open a related issue as changing it will case other breaking changes.

src/components/Checkbox/Checkbox.tsx
duranmla: Got it @lukahartwig . Yes, @div-Leo let's keep the default behavior to `center` then.

docs/release-notes.mdx
```diff
@@ -8,6 +8,19 @@ import { TitleCard } from './components/TitleCard';
     Keep up to date with the latest releases of the design system.
 </TitleCard>
 
+## 1.25.2 (Unreleased)
+#### 💅 Enhancements
+- `Checkbox` small & medium variants have been introduced
+
```
lukahartwig, line 14: Can you remove this empty line?

docs/release-notes.mdx
```diff
@@ -8,6 +8,19 @@ import { TitleCard } from './components/TitleCard';
     Keep up to date with the latest releases of the design system.
 </TitleCard>
 
+## 1.25.2 (Unreleased)
```
lukahartwig, line 11: Since this adds new APIs I would consider this a minor version change to either change the title to `1.26.0` (Unreleased) or just `Unreleased`.

lukahartwig (changes requested):

lukahartwig (approved):
LGTM

src/components/Checkbox/Checkbox.tsx
agonzalezUX: cool from my side :)

PR #46, opened 2021-04-23
Fix contributor count

lukahartwig:


duranmla (approved):
This means the pull request template doesn't work?
_I did something wrong?_ 😢

lukahartwig:
> This means the pull request template doesn't work?
> _I did something wrong?_ 😢

I made the change through the UI, not sure if that was the reason maybe?

Issue #38, opened 2021-04-22
Setup semantic release and CD to being able to auto deploy when merging to master


duranmla:
We need to be able to deploy under https://www.npmjs.com/org/freenow organization

lukahartwig:
There is an open branch for this however I was struggling with making the CHANGELOG update in the documentation.

See https://github.com/freenowtech/wave/tree/feature/add-semantic-release

PR #50, opened 2021-04-23
docs(CONTRIBUTING): fix typo `graps` => `grabs`

arturmiglio:
Just a small fix of a typo, to open the "door" for more contributions :)

In `CONTRIBUTING.md`:

> Feel like doing a bit more? You can check what is up for ~graps~**grabs** in our public backlog and create a PR for it


docs/release-notes.mdx
```diff
@@ -8,6 +8,12 @@ import { TitleCard } from './components/TitleCard';
     Keep up to date with the latest releases of the design system.
 </TitleCard>
 
+## Unreleased
```
nlopin, line 11: We don't need to put this into release because it is technical documentation, not public

nlopin (dismissed):

arturmiglio:
Added another fix in the `docs/contributing` documentation page.

One small issue is that the installation instructions are not currently valid, since the package is not yet published. Or maybe it is published under a different scope with public access?

Maybe it's a good idea to note that on the docs...

duranmla:
Thanks for contributing @arturmiglio indeed we need to make a manual release while https://github.com/freenowtech/wave/issues/38 gets solved

duranmla (approved):

arturmiglio:
> Thanks for contributing @arturmiglio indeed we need to make a manual release while #38 gets solved

@duranmla does it make sense to note on the README (in case people is trying to install the package with the supplied command)?

duranmla:
@arturmiglio I don't think so, I will try to make a release manually today if it is possible.
_The current usage of the library doesn't direct me to an actual state of urgency, but it is true we need to be coherent. Apologizes_

PR #61, opened 2021-04-30
ci: add semantic-release

lukahartwig:
This PR configures the Github Action to automatically release changes to `main` branch. It also creates an automatic changelog based on the commits in the release.

### DIFF
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -10,9 +10,12 @@ on:
 
 jobs:
     build:
+        if: github.ref != 'refs/heads/main'
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v2
+              with:
+                  fetch-depth: 0
             - uses: actions/setup-node@v2
               with:
                   node-version: 14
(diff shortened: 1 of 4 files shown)


nlopin:
Am i right you propose to automatically generate changelog.md? From the commit messages?

lukahartwig:
That was my idea, yes. With this change every commit to `main` may lead to a release so updating the changelog manually seems like too much effort. Additionally he changes also get displayed in the releases tab in the repo.

The commit messages can be written while merging branches in the Github UI so we still have full control over what gets displayed there.

This is just the proposal though, I'm totally open for other approaches as well.

lukahartwig:
`GITHUB_TOKEN` is injected automatically (except on forks), `NPM_TOKEN` was added by me under Settings > Secrets.

nlopin:
> The commit messages can be written while merging branches in the Github UI so we still have full control over what gets displayed there.

Let's hope we won't forget about it. I tried to use automatic release notes but it backslid every time and created mess with automatic merge messages

nlopin (approved):

snapsnapturtle (approved):

PR #64, opened 2021-05-05
fix: isLastPage true if 0 items

div-Leo:
**What**
When the `Pagination` component gets 0 items "nextPage" and "skipToLastPage" buttons are not disabled.

**Why**
When the component gets 0 items `isLastPage` is never true, because `totalItems / pageSize` with `totalItems === 0` gives `0` and `value` (the current page) must be `>= 1`.

**How**
In the case we don't have items `!totalItems`, then `isLastPage` is set to true.

Screenshot 2021-05-05 at 11 33 21

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -105,7 +105,7 @@ const Pagination: React.FC<PaginationProps> = ({
     onSkipBackward
 }) => {
     const isFirstPage = value === 1;
-    const isLastPage = value === Math.ceil(totalItems / pageSize);
+    const isLastPage = value === Math.ceil(totalItems / pageSize) || !totalItems;
 
     return (
         <Container>
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -105,7 +105,7 @@ const Pagination: React.FC<PaginationProps> = ({
     onSkipBackward
 }) => {
     const isFirstPage = value === 1;
-    const isLastPage = value === Math.ceil(totalItems / pageSize) || !totalItems;
+    const isLastPage = totalItems > 0 && pageSize > 0 ? value === Math.ceil(totalItems / pageSize) : false;
 
     return (
         <Container>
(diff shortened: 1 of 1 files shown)


rafael-sepeda:
Do we actually need to show the pagination when there are 0 items?

div-Leo:
> Do we actually need to show the pagination when there are 0 items?

That's a fair question. So far the current behaviour is this:
Screenshot 2021-05-05 at 12 29 22

IMO it should be the user who decides not to render at all the component if there are 0 items, and the bug still should be fixed.
If we want this to be a library constraint then we should anyway change the component behaviour, returning null if there're 0 items.

snapsnapturtle:
I agree with both of you. We should fix this behaviour in the library, but also hide the pagination in the projects if there are not multiple pages! Thanks for opening the PR 👍

src/components/Pagination/Pagination.tsx
```diff
@@ -105,7 +105,7 @@ const Pagination: React.FC<PaginationProps> = ({
     onSkipBackward
 }) => {
     const isFirstPage = value === 1;
-    const isLastPage = value === Math.ceil(totalItems / pageSize);
+    const isLastPage = value === Math.ceil(totalItems / pageSize) || !totalItems;
```
snapsnapturtle, line 108: Nit: I'd suggest to make the new totalItems check a little more explicit, by either checking specifically against 0 (i.e. `totalItems === 0;`) or just defaulting to 1 (i.e. `pageSize) || 1;`).

snapsnapturtle (approved):

src/components/Pagination/Pagination.tsx
div-Leo: `totalItems === 0` would not cover cases like `null` o `undefined`.  `totalItems` should be a number type anyway so shall we be concerned about it?

`value === (Math.ceil(totalItems / pageSize) || 1);` I'm not sure is more readable, I find it less explicit then saying "is last page if we don't have total items"

Which do you this we should go for? 🙂

nlopin: What about dumb `(totalItems > 0 && pageSize > 0) ? Math.ceil(...) : false`

We allow any number, so potentially we can expect NaN and Infinity 🤷‍♂️ Let's say the behavior is not defined for that cases, but for all finite numbers I think we can check

nlopin (approved):

duranmla:
It seems we can merge this @div-Leo

Issue #71, opened 2021-05-11
Textarea uses the type HTMLInputElement rather than HTMLTextAreaElement


chrisrolfe198:
I've created a codesandbox here to better show this issue: https://codesandbox.io/s/xenodochial-carson-4njwh?file=/src/App.tsx (Note the visual aspect isn't working for some reason but the actual error is when you hover over the code to see the TS error).

Essentially because the Textarea component uses HTMLInputElement it means that it is not compatible with any types that hint at HTMLTextAreaElement.

lloydaf:
@nlopin should [this](https://github.com/freenowtech/wave/blob/main/src/components/Textarea/Textarea.tsx#L43) be [HTMLTextAreaElement](https://codesandbox.io/s/objective-cloud-i0ik4b?file=/src/App.tsx:93-165)?

renejfc:
Closing this as it seems that issue and comments were resolved.

###### Please open a new ticket if you think this still needs attention.

PR #73, opened 2021-05-11
fix: use default export instead of named export

lukahartwig:
**What:**
During the investigation of #71 we noticed that `Input` and `Textarea` don't render on codesandbox. This affects only users of the ESM version of the library.
​
**Why:**
Fixes a potentially live bug.
​
**How:**
Both components used a named import for `React` by using `import { default as React } from "react"`. I changed this to `import React from "react"`.

- [X] Ready to be merged


duranmla (approved):
Oh boy! That was fast!

PR #76, opened 2021-05-13
Allow Icons color for payments icons

### LINKED ISSUE #20: Some icons do not respect the "color" prop
Some icons do not respect the `color` prop and use their own colors instead:
- CrosshairsIcon
-  All Payment Icons
-  All Service type icons

duranmla:
**What:**
Allow all icons on payments to get custom color
​
**Why:**
Closes #20

The problem is due to the fact of clustering together brand icons like `Visa` with other payment icons like `BankNoteIcon` while the former one shouldn't allow `color` but the other should support `color` updates
​
**How:**
Remove `payments` from the `exclude-color` flag on `npm run generate`
​
**Media:**
https://www.loom.com/share/155f0dd56e9547e5ab04519171d315a6

**Checklist:**
- [x] Release notes added
- [x] Ready to be merged

_I think we should be strict with the interface definition and we should separate icons payments from icons that belongs to payments brand_


nlopin:
Following your proposal, I think we can just add `brands` folder and keep all the logos there

duranmla:
I have created a dedicated folder for brands. In that sense, I have updated the code accordingly and now the documentation looks like 
image

snapsnapturtle (approved):

nlopin (approved):

agonzalezUX (approved):

PR #77, opened 2021-05-13
Inline spinner update

### LINKED ISSUE #22: InlineSpinner in a Button component breaks alignment with other Buttons
When placing two `Button` next to each other and using the `InlineSpinner` on one of them, it is breaking the alignment, when the InlineSpinner spins. As the screen record shows, when the first buttons starts the spinner, the second button jumps down and back up when spinning ended.

```
<Button type="submit" mr={2} disabled={state.isLoading || campaignName === ''}>
  {state.isLoading && <InlineSpinner color={Colors.WHITE} />}
  {formatMessage({ id: 'button.createCampaign' })}
</Button>
<Button onClick={handleCampaignLaunch} disabled={isButtonDisabled || isCampaignLaunched} variant={variant}>
  {state.isLoading && <InlineSpinner mr={1} />}
  {children}
</Button>
```

Workaround was to put those Buttons in a `div` with `display: flex;`

https://user-images.githubusercontent.com/8927747/114019141-56803780-986e-11eb-8cf9-f6c6b48dcbc1.mov

duranmla:
**What:**
Allow `InlineSpinner` to render as other icons when placed alongside a text
​
**Why:**
Relates #22 

In some cases when the spinner is being rendered with text it breaks the layout
​
**How:**
Wrap the icon with `span` _(as changing the div with span doesn't solve the issue)_ so we can reproduce the scenario similar to https://codesandbox.io/s/button-spinner-issue-wnmgv?file=/src/App.tsx

**Checklist:**
- [x] Release notes added
- [x] Ready to be merged

**Note:**
_I am a bit hesitant with the approach but at the same time I think It will solve the issue and there is no clear drawbacks I can see_

### DIFF
--- a/src/components/InlineSpinner/InlineSpinner.tsx
+++ b/src/components/InlineSpinner/InlineSpinner.tsx
@@ -36,7 +36,7 @@ const rotation = keyframes`
     }
 `;
 
-const InlineSpinner: React.FC<InlineSpinnerProps> = styled.div<InlineSpinnerProps>`
+const InlineSpinnerIcon: React.FC<InlineSpinnerProps> = styled.div<InlineSpinnerProps>`
     display: inline-block;
     box-sizing: border-box;
     width: 1.25rem;
@@ -49,6 +49,12 @@ const InlineSpinner: React.FC<InlineSpinnerProps> = styled.div<InlineSpinnerProp
 
     ${compose(margin, sizeVariant)}
 `;
+
+const InlineSpinner = () => (
+    <span>
+        <InlineSpinnerIcon />
+    </span>
+);
 
 InlineSpinner.defaultProps = {
     color: Colors.AUTHENTIC_BLUE_900,
--- a/docs/release-notes.mdx
+++ b/docs/release-notes.mdx
@@ -15,6 +15,7 @@ import { TitleCard } from './components/TitleCard';
 
 #### 🐛 Bug Fixes
 - `Select` to have cursor pointer not-allowed
+- `InlineSpinner` to behave closer to other icons when placed alongside text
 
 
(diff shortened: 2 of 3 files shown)


src/components/InlineSpinner/InlineSpinner.tsx
```diff
@@ -50,6 +50,12 @@ const InlineSpinner: React.FC<InlineSpinnerProps> = styled.div<InlineSpinnerProp
     ${compose(margin, sizeVariant)}
 `;
 
+const InlineSpinner = () => (
+    <span>
```
nlopin, line 54: do we understand why we need this span? Semantically it looks strange `span > div`

agonzalezUX (approved):

src/components/InlineSpinner/InlineSpinner.tsx
duranmla: It seems to be related to the way inline display works. But I am not able to articulate a full response over this _(That's why I add the note I added on the PR description._ Rather I will point you to:
- https://codesandbox.io/s/button-spinner-issue-wnmgv?file=/src/App.tsx
- https://mytaxi.slack.com/archives/CCWTB4C6L/p1620738413014800

nlopin:
I asked for a person who is very good at CSS (@solarrust, hello) and the solution to the issue is to set `vertical-align: bottom` or `top` to the button without the loader.

duranmla:
Right @nlopin but that's within the thread I link to you. The thing is that this can potentially happen on all containers of this icon. _(the icon can potentially break a layout)_ so the idea will be to make the icon be easy to drop in rather than solve the issue per "host" component. Do I explain myself?

PR #78, opened 2021-05-13
Update label background for boxed input variant

### LINKED ISSUE #14: Input label has a white background on coloured backgrounds
With the current implementation, we are using a white background for the label in the active position which overlays on the border, giving the effect that the line is broken by the label. This has the undesired side effect that when the Input component is used on a background which isn't white, the label background will show up.

Untitled

duranmla:
**What:**
Allow label background to "respect" host background-color
​
**Why:**
Closes #14​

**How:**
Apply a gradient of 50%/50% that will take half of the label + _(hardcoded value check note [1])_

**Media:**
image
image
image

**Checklist:**
- [x] Release notes added
- [x] Ready to be merged

**Note:**
1. There are two variables taking place here. `fontSize`, `top` both of them change with the `size` attribute and the units themselves are odd numbers `0.375rem` or `0.75rem` this makes it harder to create a deterministic number instead of a "magic" one. The normal variant size medium will work great with `0.0625rem` matching the inset value but once the size small take place _I got confused_
2. I have a `TS` issue in regards to the prop.size in which the value is not mapped to `small` | `medium`

> Any suggestions/help to fix or improve these two points above is welcome. For now, these changes don't break anything and will solve the issue

Props to @nlopin for the gradient idea! 🚀

### DIFF
--- a/docs/release-notes.mdx
+++ b/docs/release-notes.mdx
@@ -16,6 +16,7 @@ import { TitleCard } from './components/TitleCard';
 
 #### 🐛 Bug Fixes
 - `Select` to have cursor pointer not-allowed
+- `Input` boxed variant label respect host background color
 
 
 #### 📖 Documentation
--- a/docs/release-notes.mdx
+++ b/docs/release-notes.mdx
@@ -16,7 +16,6 @@ import { TitleCard } from './components/TitleCard';
 
 #### 🐛 Bug Fixes
 - `Select` to have cursor pointer not-allowed
-- `Input` boxed variant label respect host background color
 
 
 #### 📖 Documentation
--- a/src/components/Input/BoxedInput.ts
+++ b/src/components/Input/BoxedInput.ts
@@ -47,6 +47,11 @@ const BoxedInput: FC<InternalInputComponentProps> = styled(BaseInput)`
         ${p => (p.hasValue || p.placeholder ? activeBoxedPosition(p.size) : null)};
         color: ${getLabelColor};
         background: ${p => (p.inverted ? Colors.AUTHENTIC_BLUE_900 : Colors.WHITE)};
+        background: ${p =>
+            `linear-gradient(0deg, ${p.inverted ? Colors.AUTHENTIC_BLUE_900 : Colors.WHITE} calc(50% + ${
+                // @ts-ignore
+                p.size === 'small' ? '0.0825rem' : '0.0625rem'
+            }), transparent 50%)`};
     }
 
     ${p => (p.error ? errorStyles : null)}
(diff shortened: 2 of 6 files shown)


docs/release-notes.mdx
```diff
@@ -16,6 +16,7 @@ import { TitleCard } from './components/TitleCard';
 
 #### 🐛 Bug Fixes
 - `Select` to have cursor pointer not-allowed
+- `Input` boxed variant label respect host background color
```
nlopin, line 19: is it a bug or an enhancement? 

I think the agreement is to put only the most important things into the documentation and keep everything in the changelog. Right @lukahartwig ?

nlopin (approved):

docs/release-notes.mdx
duranmla: Cool, let me know how to proceed here before merge this.

lukahartwig: If the merge commit has the prefix `fix:` then it will show up as bug fix in the changelog. I think we should use the release notes only for important stuff like features or breaking changes.

lukahartwig (approved):
Really nice solution!

src/components/Input/BoxedInput.ts
```diff
@@ -47,6 +47,11 @@ const BoxedInput: FC<InternalInputComponentProps> = styled(BaseInput)`
         ${p => (p.hasValue || p.placeholder ? activeBoxedPosition(p.size) : null)};
         color: ${getLabelColor};
         background: ${p => (p.inverted ? Colors.AUTHENTIC_BLUE_900 : Colors.WHITE)};
+        background: ${p =>
+            `linear-gradient(0deg, ${p.inverted ? Colors.AUTHENTIC_BLUE_900 : Colors.WHITE} calc(50% + ${
+                // @ts-ignore
+                p.size === 'small' ? '0.0825rem' : '0.0625rem'
+            }), transparent 50%)`};
```
div-Leo, line 54: I think the `background` property above (line 49) is going to be overwritten by this one and ignored.

src/components/Input/BoxedInput.ts
```diff
@@ -71,6 +76,11 @@ const BoxedInput: FC<InternalInputComponentProps> = styled(BaseInput)`
             ${p => activeBoxedPosition(p.size)};
             color: ${p => (p.inverted ? Colors.WHITE : Colors.ACTION_BLUE_900)};
             background: ${p => (p.inverted ? Colors.AUTHENTIC_BLUE_900 : Colors.WHITE)};
+            background: ${p =>
+                `linear-gradient(0deg, ${p.inverted ? Colors.AUTHENTIC_BLUE_900 : Colors.WHITE} calc(50% + ${
+                    // @ts-ignore
+                    p.size === 'small' ? '0.0825rem' : '0.0625rem'
+                }), transparent 50%)`};
```
div-Leo, line 83: Nit: Maybe it's worth considering to avoid the repetition with the above style as it's exactly the same unless we prefer to have them separate

agonzalezUX (approved):

src/components/Input/BoxedInput.ts
duranmla: Yes, that's the idea. Although we don't enforce to support legacy browsers, I found it convenient to apply these "hacky" solutions as a sort of incremental enhancement.

src/components/Input/BoxedInput.ts
duranmla: The thing is that I feel I am going to touch in a much bigger refactor. I see that previously the pattern for the background is the same _(duplication)_ and I see that we have some functions to encapsulate some styles. Perhaps, instead of reactively doing something here we may want to step back and think about a better way.

snapsnapturtle (approved):

PR #86, opened 2021-05-20
Link inputs and labels when `id` is not provided by the user

nlopin:
**What:**
Fixes critical accessibility issue for `Input`, `Textarea` and `Select`. Always links input to the corresponding labels
​
**Why:**
It is recommended to have inputs and corresponding labels linked. The input must provide an `id` and label references this id in the `for` attribute.

In the mentioned above components we create this relation only when the developer passed `id` property to the component. If not, the component seriously violates [the accessibility guidelines about the labels](https://dequeuniversity.com/rules/axe/4.2/label)

**How:**

New dependency [nanoid](https://github.com/ai/nanoid) which is 108 bytes minified and gzipped.

1. If the user provides an `id` prop we use it as an id
2. If not, but there is no label, do nothing
3. If not, and the label exists, generate a random id in the form `wave-{6 random characters)`

**Media:**

https://www.loom.com/share/ddaf8f367e4d4b7b9eb5afc830df80ca

**Checklist:**

- [x] Ready to be merged

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -1,7 +1,6 @@
-import React, { forwardRef, useEffect, useMemo, useState } from 'react';
+import React, { forwardRef, useEffect, useState } from 'react';
 import { extractClassNameProps, extractWidthProps, extractWrapperMarginProps } from '../../utils/extractProps';
-import { generateId } from '../../utils/ids';
-
+import { useGeneratedId } from '../../utils/hooks/useGeneratedId';
 import { BottomLinedInput } from './BottomLinedInput';
 import { BottomLinedInputLabel } from './BottomLinedInputLabel';
 import { BoxedInput } from './BoxedInput';
@@ -15,11 +14,7 @@ const Input = forwardRef<HTMLDivElement, InputWrapperProps & InputProps>((props,
     const { widthProps, restProps } = extractWidthProps(withoutMargin);
 
     const { label, onChange, size, ...rest } = restProps;
-    const id = useMemo(() => {
-        if (props.id) return props.id;
-
-        return label ? generateId() : null;
-    }, [props.id]);
+    const id = useGeneratedId(props.id);
 
     const [hasValue, setHasValue] = useState(rest.value && rest.value.toString().length > 0);
 
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -1,5 +1,7 @@
-import React, { forwardRef, useEffect, useState } from 'react';
+import React, { forwardRef, useEffect, useMemo, useState } from 'react';
 import { extractClassNameProps, extractWidthProps, extractWrapperMarginProps } from '../../utils/extractProps';
+import { generateId } from '../../utils/ids';
+
 import { BottomLinedInput } from './BottomLinedInput';
 import { BottomLinedInputLabel } from './BottomLinedInputLabel';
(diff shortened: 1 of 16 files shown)


duranmla:
As we discussed in the sync can we add an `id` to the document example? So we somehow "encourage" people to use the attribute? _I guess it can behave as a little hint_

src/components/Input/__snapshots__/Input.spec.tsx.snap
```diff
@@ -1,150 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Input should set the htmlFor attribute for the label 1`] = `
-.c3 {
```
duranmla, line 4: It is super great that we have removed this noise, nonetheless, what do you think about having these styles for at least one test so we can audit if any detail on CSS has been updated unconsciously?

src/components/Input/Input.tsx
```diff
@@ -12,7 +14,12 @@ const Input = forwardRef<HTMLDivElement, InputWrapperProps & InputProps>((props,
     const { marginProps, restProps: withoutMargin } = extractWrapperMarginProps(withoutClassName);
     const { widthProps, restProps } = extractWidthProps(withoutMargin);
 
-    const { label, onChange, size, id, ...rest } = restProps;
+    const { label, onChange, size, ...rest } = restProps;
+    const id = useMemo(() => {
```
duranmla, line 18: This will probably repeat across several instances of different components. Makes sense to have a `useGeneratedId(props.id)` hook that we can reuse everywhere?

duranmla (approved):
Love to see this type of improvement coming. Couple of doubts/questions but other than that LGTM!

src/components/Input/__snapshots__/Input.spec.tsx.snap
nlopin: there are a lot of snapshots still

src/components/Input/Input.tsx
nlopin: I see that the code is duplicated, but if I want to extract it, I will have to make it generic enough for reuse. The way I see the generic makes it complex because we need to provide an option to avoid id generation (see the use of `label`):

```
const useGeneratedId = (id, abortCondition) => {
  return useMemo(() => {
    if (id) return id;
    if (abortCondition()) return null

    return generateId()
  }, [id])
}
```

Feels a bit too much to me. What do you think  @duranmla @snapsnapturtle @lukahartwig ?

nlopin: The idea behind the logic is the following. 

1. The user set props have the highest priority, we cannot drop them, we suppose they know what they do. Even if `id` is not needed (e.g. there is no label set), we set it anyways.
2. We generate `id` only when it must present

agonzalezUX (approved):

src/components/Input/Input.tsx
duranmla: But it is not like:

```js
const useGeneratedId = (id) => {
  return useMemo(() => {
    if (id) return id;
    return generateId()
  }, [id])
}
```

Why do we need the "abort condition", users will set or not set the id isn't? Nevertheless, if the abstraction is expensive I will for sure favor the duplication of a simpler code instead. 👌

nlopin: Probably it is a premature optimization, I'll do the simple version and then we'll see

PR #87, opened 2021-05-20
feat: replace release notes with changelog

snapsnapturtle:
**What:**
We have removed the release notes page, containing old versions from the internal package and replaced it with the generated changelog.
​
**Why:**
With the move to Github and going open source, old versions that were mentioned in the release notes are not available in the @freenow/wave package. This reduces the complexity for adding new code to the project by not requiring the developers to add a manual entry to the release notes document for their changes.
​
**Checklist:**

- [x] Ready to be merged
      

Unfortunately this is not very testable, so we will have to cross our fingers and hope it works out 🤞


agonzalezUX (approved):

duranmla (approved):
Fingers cross then! 🤞

PR #89, opened 2021-05-21
`Password` component

### LINKED ISSUE #81: New Password input component
We would like to create a new component for inputting/creating a password. Based on a design we were provided for one of our projects.
Screenshot 2021-05-17 at 12 42 02
Screenshot 2021-05-17 at 12 42 13

The difference from the already existing <Input /> of a type password is the eye icon which enables user to switch between the visibility of the inputted text.

Technical solution:
We would reuse the existing Input component, with the fixed password type and adding an eye icon which will work as a button switching between input type -> password/text

Options:
1) Create a specific Input type - being included in the Input folder ( + it's kinda a derivate from already existing Input anyway)
2) Create a stand alone component ( + better visibility)

nlopin:
**What:**
New `Password` component to log in or set passwords.

It has two modes `hidden` (default, the password is hidden) and `shown` (the password is shown as plain text). A user can toggle between modes with an eye button.

Kudos to @dani-walnut for help
​
**Why:**
Closes #81 

It is needed across our login pages for passengers, drivers, and companies. Plus for reset password web pages.
​
**How:**

The component is a stateful wrapper of the existing `Input`.

The approach to the component's accessibility is inspired by a [gov.uk recent article about their implementation of the password component].
​
**Media:**

The component page:
Screen Shot 2021-05-21 at 22 40 02-fullpage

♿️ The accessibility demo

[Video demo](https://www.loom.com/share/08aee2826dc44b92a7e785da3446a933)

**Checklist:**

- [x] Add `aria-controls` to the button when #86 is merged
- [x] Ready to be merged


nlopin:
I also added the `VisuallyHidden` component to be able to hide an element (usually, text) visually, but keep it visible for screen readers. This is a widely used pattern

src/components/Password/ToggleButton.ts
```diff
@@ -0,0 +1,47 @@
...
+        return p.variant === 'bottom-lined' ? '.5rem' : 0;
+    }};
+    ${inputType};
+
+    svg path {
+        fill: var(--color);
+    }
+
+    &:hover svg path {
+        fill: var(--hover-color);
```
nlopin, line 43: This is the only way the svg reacted properly to the hover. The regular `${p => ...}` didn't work. 

I actually like this approach more because it is pure CSS + we don't know anything about the color inside

src/components/Password/Password.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+const ControlledPassword = props => {
+    const [value, setValue] = useState('');
+    const handleChange = e => {
+        setValue(e.target.value);
+    };
+
+    return <Password {...props} id="password-id" value={value} onChange={handleChange} />;
+};
+
+describe('Password', () => {
```
duranmla, line 16: How do we know this will work properly when using like:

```js
<form>
    <Password name="password" />
    <button type="submit">Submit</button>
</form>
```

I'd expect on submitting the form I'd be able to get the value of the password.  `new FormData(form)` as we are expected from any input. What are your thoughts on validating that? 

https://codesandbox.io/s/wave-form-s83d3?file=/src/App.tsx

> It probably works as we are using `Input` component under the hood

src/components/Password/Password.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { ToggleButton } from './ToggleButton';
+import { useGeneratedId } from '../../utils/hooks/useGeneratedId';
+
+const PasswordWrapper = styled.div`
+    display: inline-block;
+    position: relative;
+    box-sizing: border-box;
+
+    input {
+        padding-right: 2.5rem;
```
duranmla, line 19: Maybe worth adding a comment to know this is due to the icon?

src/components/Password/ToggleButton.ts
```diff
@@ -0,0 +1,47 @@
...
+    background: transparent;
+    border-radius: ${get('radii.2')};
+    border: none;
+    cursor: pointer;
+    display: inline-flex;
+    justify-content: center;
+    text-align: center;
+    text-decoration: none;
+    height: 100%;
+    width: 2.5rem;
```
duranmla, line 28: Shouldn't we try to relate this to the padding 2.5rem we have within `PasswordWrapper`, perhaps a constant file?

duranmla (approved):
There are a couple of comments/doubts/suggestions. Either way, I feel you folks have done extraordinary work. I love the a11y section and the video showing the VoiceOver, also I think the Advice component is super helpful.

Thanks a lot for going the extra mile with all the effort here, love to see this awesome work!

Really nice work @nlopin @dani-walnut let's keep rocking!

duranmla:
Hey BTW, this can be a bit of a blocker but seeing your screenshot I think that the small variant is not properly vertically align with the input _(at least is the perception I have with the icon on the right side)_. Can we take a look into that before merge?

duranmla (commented):
I thought we agree to have this component rendered whenever `Input type=password` as defined https://github.com/freenowtech/wave/issues/81#issuecomment-842890657

src/components/Password/docs/Password.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+### Sign up or reset password
+
+If you are in the situation when a user needs to create a new password, set `purpose` prop to `new-password`:
+
+```jsx
+<Password purpose="new-password" label="New password" />
+```
+
+This will allow password managers to autogenerate the password for the user.
```
snapsnapturtle, line 40: > to generate a password for the user automatically.

src/components/Password/ToggleButton.ts
snapsnapturtle: I was also thinking about introducing more css variables. They seem to become more and more widespread

src/components/Password/ToggleButton.ts
```diff
@@ -0,0 +1,47 @@
...
+    right: 0;
+    align-items: ${p => (p.size === 'small' || p.variant === 'bottom-lined' ? 'end' : 'center')};
+    padding-bottom: ${p => {
+        if (p.size === 'small') return '.25rem';
```
snapsnapturtle, line 32: Let's prefix this and the next line with a `0`, so it follows what we set up with prettier

nlopin:
> I thought we agree to have this component rendered whenever `Input type=password` as defined [#81 (comment)](https://github.com/freenowtech/wave/issues/81#issuecomment-842890657)

We use `<Input type="password" />` internally in the current solution so it is not possible in the current setup. If we want to have this alias, I will have to create a wrapper around the input and render `Password` or `InternalInput` component depending on the `type` passed. 

Do we still want to do so and leave our users without the basic HTML password option?

nlopin:
> Hey BTW, this can be a bit of a blocker but seeing your screenshot I think that the small variant is not properly vertically align with the input _(at least is the perception I have with the icon on the right side)_. Can we take a look into that before merge?

@agonzalezUX @rafael-sepeda should I move it a couple of pixels down?

src/components/Password/ToggleButton.ts
nlopin: Done

snapsnapturtle:
> Do we still want to do so and leave our users without the basic HTML password option?

In my opinion we should have two components that can be used to show passwords. A) the `<Input type="password" />` one, where we don't have the eye toggle and B) `<PasswordInput />` which uses the input. It reduces complexity in the library and allows for more flexibility when using the library

src/components/Password/Password.spec.tsx
nlopin: We can validate this, but technically we will validate HTML spec 🤷  If you think it is valuable I don't see any issues adding this check

duranmla:
In my side, I feel the Design system is made to make opinions and unify UI/UX, I don't think we should offer the flexibility but rather enforce the usage of the Password input we develop _(I see it like "this is the way a input password looks in our ecosystem")_

snapsnapturtle (approved):

agonzalezUX:
I think everything is good.
However, I have the feeling that in the default input the icon pops a bit too much, I know this is how its designed but maybe we could improve it, what do you think @rafael-sepeda ?

And also, in the disabled version I am not sure if the eye should also be rendered.... what do you think?

for the rest super <3

nlopin:
> And also, in the disabled version I am not sure if the eye should also be rendered

At the moment, it is hidden:
image

nlopin:
About the `type=password` discussion in favor of Alexis I can say that we gain more by going this way. At the moment people update the library, they get the new password component with the eye and a11y in place.

IDK if we consider this a breaking change or not, though

nlopin:
I've tried to keep the best from both worlds:
* if a user does `<Input type="password" />`, we render the `<Password />` component
* there is still a way to create a vanilla password input via `<InnerInput type="password" />`

agonzalezUX (approved):

PR #96, opened 2021-05-27
improvement: adding the disabled text color, fixing inverted text colors

lloydaf:
**What:**
This PR adds the `disabled` state to the `Text` component, in alignment with the design for wave. It also fixes colors when `inverted` (for `weak` and `disabled` texts) also in alignment with the design.
​
**Why:**
This is in regard to [SUP-11112](https://jira.intapps.it/browse/SUP-11112)
​
**How:**
* Add a new disabled property to the `Text` component
* Fix the colors for inverted state
* Add examples to the documentation
​
**Media:**
Design
Documentation Screenshot

PR #97, opened 2021-05-27
Adds disabled checkbox text color

lloydaf:
**What:**
This PR fixes the color of the text of the `disabled` state for the `Checbox` component.
​
**Why:**
This is in regard to [SUP-11111](https://jira.intapps.it/browse/SUP-11111)
​
**How:**
* Pass the `disabled` property to the `Text` component inside `Checkbox`
​
**Media:**
Design
Screenshot of documentation

> Note: This PR should ideally be merged only after [the preceding one](https://github.com/freenowtech/wave/pull/96) is merged, so that we can isolate changes in the merge commit with the ticket (currently, changes for [SUP-11112](https://jira.intapps.it/browse/SUP-11112) are included here as well)

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -49,7 +49,11 @@ const Checkbox: FC<CheckboxProps> = props => {
     let dynamicLabel: ReactNode = label;
 
     if (typeof label === 'string') {
-        dynamicLabel = <Text fontSize={size}>{label}</Text>;
+        dynamicLabel = (
+            <Text disabled={disabled} fontSize={size}>
+                {label}
+            </Text>
+        );
     }
 
     return (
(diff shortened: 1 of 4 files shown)


duranmla:
Now I understand https://github.com/freenowtech/wave/pull/96 nonetheless, I am not sure if we should achieve this by using the parent _(although I understand it is not a "recommended" practice on styled-components)_.

Kinda,
```js
Checkbox = css`
    ${Text} {
        // disable styles
    }
`
```

But at the same time _Why not allow disabled prop on `Text`_. I am a bit hesitant right now, I would like to read some other folks comments

lloydaf:
In my opinion, since we're adding a disabled property to `Text`, this is the best way to ensure consistency. Tomorrow, if we decide to change the disabled colors for example, we only have to change it in `Text` and it will be reflected everywhere else. So in my opinion this is also easier from a maintainability perspective, but I am open to hearing different opinions as well.

snapsnapturtle (approved):

duranmla:
@all-contributors please add @lloydaf for code collaboration after this pull request

PR #102, opened 2021-06-02
Use correct version of wave in package.json

nlopin:
**What:**
Fixes the version of the library in package.json
​
**Why:**
I don't know why but the version in package.json is out of sync. It probably happened when we moved it from the internal repo.

**Checklist:**

- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@freenow/wave",
-    "version": "1.25.1",
+    "version": "1.1.0",
     "description": "React components of the Wave design system for your Front-End project",
     "main": "lib/cjs/index.js",
     "typings": "lib/types/index.d.ts",


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@freenow/wave",
-    "version": "1.25.1",
+    "version": "1.1.0",
```
snapsnapturtle, line 3: Didn't we just release version 1.10.0?

snapsnapturtle (approved):

duranmla:
I tend to feel this should be done like this. When going to https://semantic-release.gitbook.io/semantic-release/support/faq#why-is-the-package-jsons-version-not-updated-in-my-repository you can see that 

>  _@semantic-release/git plugin can be used to push the updated package.json as well as other files to the Git repository_

Maybe that's the actual change we want to do.

agonzalezUX (approved):

duranmla:
I think we should close this pull request and wait for a more thoughtful implementation using semantic release.
Has this hasn't been merge yet I guess we are agree. @nlopin

nlopin:
I've created a separate issue #119 to address this separately

PR #105, opened 2021-06-04
Separates headline styles from heading level

### LINKED ISSUE #17: Update style and semantics of Headline component
As we currently set the headline style based on the underlying component (h1 to h6), it's impossible to set a different semantic headline level from the visual style. We want to be able to set the visual style independently of the underlying semantic headline level.

For this, we will export all six headlines as individual components and allow them to configure the "as" prop to override the default headline level.

My suggestion for the new names of exported headlines (from h1 to h6) is:
- HeadlineXXLarge
- HeadlineXLarge
- HeadlineLarge
- HeadlineMedium
- HeadlineSmall
- HeadlineXSmall

With these changes it will be possible to use a smaller headline with the semantic largest and most prominent headline like this `<HeadlineMedium as="h1"/>`.

nlopin:
**What:**
Allow influencing the size without changing the headline level.
​
**Why:**
Closes #17 

**How:**
Expose the `size` prop which takes precedence over the default headline level style
​
**Checklist:**

- [x] Ready to be merged
      

This is not a breaking change. The components without `size` prop set will still behave as in previous versions

### DIFF
--- a/src/components/Headline/docs/Headline.mdx
+++ b/src/components/Headline/docs/Headline.mdx
@@ -7,19 +7,29 @@ route: /components/headline
 import { Playground } from 'docz';
 import { ItemWrapper } from '../../../../docs/components/ItemWrapper.ts';
 import { Headline } from '../Headline.tsx';
-import { theme } from '../../../essentials/theme.ts';
-import { Colors } from '../../../essentials/Colors/Colors.ts';
+import { HeadlinePropsTable } from './HeadlinePropsTable';
+
 
 # Headline
 
-The Headline component will render an html `h1-6` tag. Please reference the
+The Headline component will render a html `h1-6` tag. Please reference the
 [w3 recommendations for headings](https://www.w3.org/WAI/tutorials/page-structure/headings/) to ensure your headings
 provide an accessible experience for screen reader users.
 
+Prefer to use headlines starting from level one and not skip levels. Use combination of `as` and `size` property to keep
+the semantics while following the design.
+
 -   Use different headlines and paragraphs to structure your content hierarchically
--   Avoid using headline lengths, breaking more then two times (three times on mobile)
+-   Avoid using headline lengths, breaking more than two times (three times on mobile)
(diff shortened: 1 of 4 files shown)


src/components/Headline/docs/Headline.mdx
```diff
@@ -7,20 +7,30 @@ route: /components/headline
...
 import { Headline } from '../Headline.tsx';
-import { theme } from '../../../essentials/theme.ts';
-import { Colors } from '../../../essentials/Colors/Colors.ts';
+import { HeadlinePropsTable } from './HeadlinePropsTable';
+
 
 # Headline
 
-The Headline component will render an html `h1-6` tag. Please reference the
+The Headline component will render a html `h1-6` tag. Please reference the
```
snapsnapturtle, line 15: I think `an` in this case is actually correct, but I'm not a native speaker either

src/components/Headline/docs/HeadlinePropsTable.tsx
```diff
@@ -0,0 +1,19 @@
...
+    const props = [
+        {
+            name: 'as',
+            type: '"h1" | "h2" | "h3" | "h4" | "h5" | "h6"',
+            defaultValue: 'h1'
+        },
+        {
+            name: 'size',
+            type: '"xxl" | "xl" | "l" | "m" | "s" | "xs"',
+            defaultValue: `depends on the heading level`
```
snapsnapturtle, line 14: We don't set a default here, this might be confusing for the user. How about we add this note in the `description` field, that the size prop will depend on the headling "as" level.

src/components/Headline/docs/Headline.mdx
```diff
@@ -54,4 +64,5 @@ provide an accessible experience for screen reader users.
 
 <Playground>
```
snapsnapturtle, line 65: While we're at it, can we change the headline for this section to `## Examples` instead of playground, like with our other documentation pages?

snapsnapturtle (approved):

Issue #95, opened 2021-05-27
InlineSpinner does not reflect the `size` prop


nlopin:
Both sizes look the same:

![image](https://user-images.githubusercontent.com/1469636/119808639-9c5b9280-bee4-11eb-9d20-b5be5ac2d730.png)

snapsnapturtle:
For reference: The small size should be `1rem` and the medium size should be `1.25rem` in width & height.

snapsnapturtle:
I noticed another issue with the inline spinner component. Since the change in pr #77, the component accepts no props anymore. This results in the margin and the size props being ignored (breaking builds in ts).

PR #107, opened 2021-06-08
fix: allow props to change the inline spinner again

snapsnapturtle:
**What:**
Re-enable all props for the InlineSpinner component and fixed the tests, so we don't break it again 👀 
​
**Why:**
We introduced a wrapper element for the InlineSpinner in #77, but didn't forward the props for margin and size. This resulted in typescript projects that set the margin to break. All other projects would simply lose their set margins and sizes. @nlopin already found the issue with the sizes in #95, which this PR closes as well.
​
**How:**
Instead of wrapping the `div` element with `span`, we set the original element to `span`.
​
**Checklist:**
- [x] Ready to be merged

### DIFF
--- a/src/components/InlineSpinner/__snapshots__/InlineSpinner.spec.tsx.snap
+++ b/src/components/InlineSpinner/__snapshots__/InlineSpinner.spec.tsx.snap
@@ -17,10 +17,12 @@ exports[`InlineSpinner renders in a custom color 1`] = `
   border-width: 0.125rem;
 }
 
-<span
-  class="c0"
-  color="#069D4F"
-/>
+<span>
+  <div
+    class="c0"
+    color="#069D4F"
+  />
+</span>
 `;
 
 exports[`InlineSpinner renders the default variant 1`] = `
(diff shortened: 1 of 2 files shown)


nlopin:
Have you checked the new approach works in the [example Alexis reported](https://codesandbox.io/s/button-spinner-issue-wnmgv)?

snapsnapturtle:
Yes @nlopin, I did check it with the example. I was hoping to be able to write a test for it, but without visual comparisons it's very difficult to have this testable

src/components/InlineSpinner/__snapshots__/InlineSpinner.spec.tsx.snap
```diff
@@ -7,16 +7,20 @@ exports[`InlineSpinner renders in a custom color 1`] = `
   width: 1.25rem;
   height: 1.25rem;
   vertical-align: text-bottom;
-  border: 0.125rem solid;
+  border: 0.125rem solid #069D4F;
```
nlopin, line 10: from where this color comes from? shouldn't it be `inherit`?

nlopin (approved):

src/components/InlineSpinner/__snapshots__/InlineSpinner.spec.tsx.snap
snapsnapturtle: This is the test to ensure the spinner accepts a custom color 🎨

PR #109, opened 2021-06-08
feat: add new textVerticalAlign prop to radio button

snapsnapturtle:
**What:**
Adding a complement to the `textVerticalAlign` prop of the Checkbox to the RadioButton component.
​
**Why:**
In one of our tools, we have a design that has a multi line radio button label, with the radio button icon aligned at the top. In addition, this increases feature parity with the Checkbox component.
​​
**Media:**
image

**Checklist:**
- [x] Ready to be merged

While working on the RadioButton component, I also updated the documentation page to follow our current standards.


src/components/RadioButton/RadioButton.spec.tsx
```diff
@@ -26,4 +26,8 @@ describe('Checkbox', () => {
     it('renders the disabled checked style', () => {
         expect(render(<RadioButton disabled defaultChecked />).container.firstChild).toMatchSnapshot();
     });
+
+    it('renders the correct text vertical align version', () => {
+        expect(render(<RadioButton label="Test" textVerticalAlign="top" />).container.firstChild).toMatchSnapshot();
```
nlopin, line 31: do you think we can check the style directly instead of matching the snapshot? I feel we overuse the snapshot testing, it is very hard to identify errors with those

snapsnapturtle: I was thinking the same thing, but we would just be verifying the implementation (check that child component a has the css variable x and so on) and not the feature. I'd prefer to dig a bit deeper into the visual comparison tests.

nlopin (approved):

PR #114, opened 2021-06-10
Feat: side (left/right) modal

div-Leo:
**What:**
Add a side (left/right) modal card.
​
**Why:**
We need a side modal for this [design](https://app.zeplin.io/project/57fe438f81b870887cc71736/screen/60900157d95163088c092d08).
​
**How:**
Add a new component `SideCard` which renders when it receives the side props ("left" | "right") and slides from the relative side.
​
**Media:**
wVGHMtm3UL

See more in this video on [loom](https://www.loom.com/share/2a8bca6293a6499ba79f8cc25913e5c1)

**Checklist:**

- [ ] Ready to be merged

### DIFF
--- a/src/components/Modal/Modal.spec.tsx
+++ b/src/components/Modal/Modal.spec.tsx
@@ -12,6 +12,14 @@ describe('Modal', () => {
         expect(render(<Modal fullscreen>Content</Modal>).container).toMatchSnapshot();
     });
 
+    it('renders the left side variation', () => {
+        expect(render(<Modal side="left">Content</Modal>).container).toMatchSnapshot();
+    });
+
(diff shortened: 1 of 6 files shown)


agonzalezUX:
I kind of like the solution, however I am not sure this right search is a modal tho, maybe we need to create a side bar component instead. What do you think @rafael-sepeda Vini?

rafael-sepeda:
@agonzalezUX agree. It works as a sidebar. And then we need something generic: a sidebar container-component, that will allow to inject different content: menu, dialogs, etc.

agonzalezUX:
maybe we can treat it as an off-canvas component
https://getbootstrap.com/docs/5.0/components/offcanvas/

PR #118, opened 2021-06-14
fix: export Offcanvas component

div-Leo:
**What:**
Export the `Offcanvas` component.
​
**Why:**
It was not exported from the main entry file
​
**How:**
Add export in the main `index.js`


duranmla (approved):
Sorry, I miss this one on the PR review. Thanks for the quick addition.

PR #125, opened 2021-06-23
feat: add new label variants and filled prop

### LINKED ISSUE #2: Add and change variants for the Label component
In order to provide more flexibility and cover more use-cases as a developer/designer I need to have more label-variants. We need to have 5 main types ("variants" now in wave) of labels:
- Default (current Neutral)
- Info (New)
- Success (current Positive)
- Warning (New)
- Danger (current Negative in design and Danger in code)

In addition to our current outlined-style we need to create an additional style-group: *Filled*. Both Outlined and Filled styles need to have 5 label types each. This story also includes the rename of the variant "info" to "neutral".

Design: https://zpl.io/aMWQ1Mr

snapsnapturtle:
**What:**
Adding new label color "info" and "warning" in addition to the ability to give all labels a more prominent style with the "filled" prop. Closes #2 .
​
**Why:**
Some of the recent designs include additional label styles that are used in tables across our tools. Currently they are Label components with custom styles. To provide a common and reusable pattern, we want to include these styles in the library.
​​
**Media:**
image

**Checklist:**
- [ ] Ready to be merged

**Other:**
This will change the styling of the variant "info" from grey to blue, while introducing a new variant (default) to replace the grey styles. The change can lead to unexpected colours of the label when the variant is explicitly specified. It will not cause any issue when no variant is specified, as the default looks identical.

### DIFF
--- a/src/components/Label/Label.spec.tsx
+++ b/src/components/Label/Label.spec.tsx
@@ -20,6 +20,14 @@ describe('Label', () => {
         `);
     });
 
+    it('has the correct info variant colors', () => {
+        expect(render(<Label variant="info" />).container.firstChild).toHaveStyle(`
+            color: ${Colors.ACTION_BLUE_900};
+            border-color: ${Colors.ACTION_BLUE_350};
+            background-color: ${Colors.ACTION_BLUE_50};
+        `);
+    });
+
     it('has the correct success variant colors', () => {
         expect(render(<Label variant="success" />).container.firstChild).toHaveStyle(`
             color: ${Colors.POSITIVE_GREEN_1000};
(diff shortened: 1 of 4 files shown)


duranmla (approved):
Thanks! 🙌

rafael-sepeda (approved):

rafael-sepeda:
Thanks!

src/components/Label/Label.spec.tsx
```diff
@@ -35,4 +43,12 @@ describe('Label', () => {
             background-color: ${Colors.NEGATIVE_ORANGE_50};
         `);
     });
+
+    it('uses the filled version if set', () => {
+        expect(render(<Label variant="success" filled />).container.firstChild).toHaveStyle(`
+            color: ${Colors.WHITE};
+            border-color: ${Colors.POSITIVE_GREEN_900};
+            backgroundColor: ${Colors.POSITIVE_GREEN_900};
```
bluessista, line 51: This should be `background-color`.

snapsnapturtle: Sorry that it's in a different PR :( https://github.com/freenowtech/wave/pull/138

Issue #121, opened 2021-06-17
Introduce semantic color variables


snapsnapturtle:
To keep the design system and the designs in Figma in sync as well as maintaining a clear structure of which colors are used where, we want to add semantic names for colors instead of the actual color names.

## Current Situation
We export the colors by their names (i.e. AUTHENTIC_BLUE_900) and use these variables in the components as well.

## Proposal

Multiple steps will be necessary to transition from the color names to semantic names, starting with adding additional exports for the new semantic names of colors (i.e. text-primary). These colors can then be used, and should be encouraged to, by the projects. In the next step, the new semantic colors are used in the components of the design system, slowly removing the actual names. In the last step we remove the existing actual names, which will result in a breaking change.

## Plan

- [x] #128 
- [x] #202
- [x] #203
- [x] #204
- [ ] #207
- [ ] #286

rafael-sepeda:
Usually the color tokens in various designs systems are not semantic and is just manages brand colors (red, grey, etc.). But some modern systems try to define several levels of abstraction for tokens (for example [Orbit](https://orbit.kiwi/design-tokens/)):

1. **Values** - These are the raw values in code such as color values (HEX, rgba), absolute and relative length units, percentages, numbers, etc.
2. **Base Variables** - Base variables are the primitive values in a design language, represented by context-agnostic names. These can be directly used, and are inherited by all other token types (think “purple-500”).
3. **Common Tokens** - Common tokens relate to a specific context or abstraction. Common tokens help communicate the intended purpose of a token and are effective when a value with a single intent will appear in multiple places (think “color-action”).
4. **Component Tokens** - Component tokens relate to a specific component using them. Their scope only targets the component they belong to. Component tokens communicate not only the targeted properties but also their states.

So we basically have 2 core levels:
- **Options** - that represent the actual values
- **Design Decisions** - How to apply options (like $color-neutral-20) to contexts (like a conventional dark background color).

rafael-sepeda:
If you think that we have too many tokens - we can try to make some of them more generic.

nlopin:
I've updated the initiative and added steps we need to follow to fully migrate to semantic colors

martimalek:
I've added a new step regarding #207

div-Leo:
Provide semantic colors also as props and not just inside styled-components functions.
Internal discussion: [message](https://mytaxi.slack.com/archives/CLM50RECU/p1650447486148809)

nlopin:
[...]

PR #128, opened 2021-06-23
Introducing semantic/functional color variables

snapsnapturtle:
Hello everyone! We've been talking about semantic colors (or functional colors) for a long time already. Recently I had some time available and started implementing a first draft, which turned into this PR. Today I am introducing the basic semantic colors, the globals and specifics for the two button components `Button` and `TextButton`. See #121. 

**What**
Implemented semantic/functional color variables from the designers in the two button components `Button` and `TextButton` following the best practices picked up in other design systems like Github's Primer. In addition we are exposing the `themeGet` function to access theme variables outside of the design system components. This way we can already set a default theme and in the future have laid the groundwork for a separate "dark" theme.

**Why**
We want to be increase alignment with the design team (which is already using semantic names for colors) and developers by providing them an easy and understandable way to find the right color to use in their project when working on components, that are not part of the design system already.  We want to expose purposeful color names that represent their purpose instead of a descriptive name (like AUTHENTIC_BLUE_900). By implementing these new color names in our own components first, we will get a good understanding and have examples to share with the devs on when they can use which color.
​
**How**
In short we are adding a new variable to the theme with purpose-based names for all the colors used in the design system called `semanticColors`. This PR updates the button components to use the method to get the newly introduced colors from the `theme.ts`. After the changes, the visual result is exactly the same, but under the hood the components will not directly access the `Colors` variables anymore. In addition we are exposing the `themeGet` function that's directly accessing the `theme.js` to get the correct color from the library. It will be able to get the color from a different theme, if we add a "dark" theme in the future.

**Next Steps**
In the near future we should update our remaining components to use semantic color names and expose those via the `semanticColors` object as well. Before the the next major release, we should deprecate the Colors object in the theme.js and as an export and recommend teams to use the `themeGet` with the new semantic colors. With the next major release, we will remove the Colors object and rename `semanticColors` to just `colors` (we can probably provide a codemod, too).

Thanks as well for the alignment with design thanks to @rafael-sepeda 🚀


lukahartwig (approved):
LGTM

rafael-sepeda (changes requested):
@snapsnapturtle 
I found that I made a mistake in the original JSON file that I share with you: `$background.warning` should use `$colors.brand.attentionYellow.50`
 (in the file it was 350)

rafael-sepeda:
Question: do you think we can simplify `borderColor` to just `border`? Since other color slots do not have the “Color” pre/post-fix?

rafael-sepeda:
@snapsnapturtle Everything looks good.
I was only not able to find the colors for the **icons of the buttons**. Or I’m missing something?

duranmla (approved):
Thanks for the hard work! 🚀

snapsnapturtle:
@rafael-sepeda regarding different colors for the icons. I missed them in the first place, but now I think: Is it necessary to have a different color for the button icons?

rafael-sepeda:
@snapsnapturtle If we have dedicated colors for text in buttons - we should have them for icons. To avoid that I can try to map text/icons to global colors (since now we have them in the new setup). I tried to do that already, but I found that because of many states its looks not very semantic/logical sometimes. If this dues not sound like a problem - we can do that. Otherwise if we think that we should have dedicated text colors - lets have them for icons (even if icon/text use the same color).

rafael-sepeda:
@snapsnapturtle Thanks a lot for the great job and for pushing this topic forward. Really cool to finally have semantic colors live, especially in the documentation 📚. The search is really great and will help a lot!

Will it be possible for more transparency to have in the color table in addition to the hex-values the brand-color references? I mean:
- `border.primary` = `Colors.AUTHENTIC_BLUE_200`

snapsnapturtle:
@rafael-sepeda absolutely, as long as the color matches the brand color exactly that should also be easily doable with a simple map. Can you create a new issue for this and add the documentation label? 🚀

PR #133, opened 2021-06-28
improvement: adding a prop called variant for datepicker range

lloydaf:
**What:** This PR intends to add a `variant` property to the datepicker

​
**Why:** This is introduced to enable developers to decide whether to show only a single month or two months, according to their needs, without breaking the current responsive behavior.

​
**How:**
1. Create a prop `variant`

​
**Media:** https://www.loom.com/share/9cf8700746d14419a0c4610face30006

**Checklist:**

- [x] Ready to be merged

### DIFF
--- a/src/components/Datepicker/DatepickerRangeInput.tsx
+++ b/src/components/Datepicker/DatepickerRangeInput.tsx
@@ -162,6 +162,12 @@ interface DatepickerRangeInputProps extends MarginProps, WidthProps {
      * The id to be assigned to the end date input
      */
     endInputId?: string;
+    /**
+     * Determines the variant
+     * @value `'compact'` displays only a single month
+     * @default 'normal'
+     */
+    variant?: 'compact' | 'normal';
 }
 
 interface DateRangeInputText {
@@ -210,6 +216,7 @@ const DatepickerRangeInput = ({
     errorHandler,
     startInputId,
     endInputId,
+    variant = 'normal',
     ...rest
 }: DatepickerRangeInputProps) => {
     const localeObject = useLocaleObject(locale);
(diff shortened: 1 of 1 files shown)


duranmla (approved):
Love it. It is simple.
Nonetheless, I will create an issue for the future in order to explore https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries

src/components/Datepicker/DatepickerRangeInput.tsx
```diff
@@ -162,6 +162,12 @@ interface DatepickerRangeInputProps extends MarginProps, WidthProps {
      * The id to be assigned to the end date input
      */
     endInputId?: string;
+    /**
+     * Determines the variant
+     * @value `'compact'` displays only a single month
+     * @default 'normal'
+     */
+    variant?: 'compact' | 'normal';
```
snapsnapturtle, line 170: Usually `variant` describes the difference between a bottom-lined and a boxed input (at the moment we only have boxed date picker versions). Have you considered exposing the `numberOfMonths` field instead?

src/components/Datepicker/DatepickerRangeInput.tsx
```diff
@@ -369,7 +376,7 @@ const DatepickerRangeInput = ({
                     <Datepicker
                         ref={ref}
                         // TODO: refer to https://stash.intapps.it/projects/DS/repos/wave/pull-requests/104/overview?commentId=168382
-                        numberOfMonths={window.innerWidth >= 768 ? 2 : 1}
+                        numberOfMonths={variant === 'normal' && window.innerWidth >= 768 ? 2 : 1}
```
snapsnapturtle, line 379: In case we go for this, we should also add a test please

src/components/Datepicker/DatepickerRangeInput.tsx
lloydaf: I thought about it, but it seemed more technical and on the path of being a workaround rather than a solution. Plus, that opens up the pandoras box of being able to have 3 months displayed, 4 months displayed etc, which will require changes in the component logic to not break anything, imo.

If you are strongly opinionated about this, maybe we can expose something like `displaySingleMonth: boolean` or other alternatives, but `compact` seemed much more developer friendly :) 

> I know the PR is merged already, but happy to keep the conversation going.

PR #137, opened 2021-06-29
Add table of contents to all documentation pages

snapsnapturtle:
**What:**
Adding a new table of content section to the documentation pages will help users find the information more quickly. The table of contents will be hidden on screens smaller than "medium".
​
**Why:**
Since some of the documentation pages are getting longer and longer, as we are adding more useful information, having a section that summarises the headlines should help users navigate more quickly. In addition this will also help us, the developers, be more consistent with the headlines we are using and more consistent when to use which one. It also allows linking to a specific section of the page directly.

**How:**
I have updated the custom template to include the headings that are provided by docz in a special section with links to the sections on the main document.
​
**Media:**

image


duranmla (approved):
Oh, mate! This is glorious! 🙏

duranmla (approved):

rafael-sepeda (approved):
This is a great idea!

lukahartwig (approved):
LGTM

PR #138, opened 2021-06-29
Fixes the spelling of the css property in the test

snapsnapturtle:
Since the PR #125 was merged a little bit prematurely, I haven't had a chance to address the code review comments. This fixes the wrong spelling of the css property when testing for a specific Label version.

### DIFF
--- a/src/components/Label/Label.spec.tsx
+++ b/src/components/Label/Label.spec.tsx
@@ -48,7 +48,7 @@ describe('Label', () => {
         expect(render(<Label variant="success" filled />).container.firstChild).toHaveStyle(`
             color: ${Colors.WHITE};
             border-color: ${Colors.POSITIVE_GREEN_900};
-            backgroundColor: ${Colors.POSITIVE_GREEN_900};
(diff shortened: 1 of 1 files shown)


duranmla (approved):
😅

src/components/Label/Label.spec.tsx
```diff
@@ -48,7 +48,7 @@ describe('Label', () => {
         expect(render(<Label variant="success" filled />).container.firstChild).toHaveStyle(`
             color: ${Colors.WHITE};
             border-color: ${Colors.POSITIVE_GREEN_900};
-            backgroundColor: ${Colors.POSITIVE_GREEN_900};
```
lukahartwig, line 51: Was this merged with a failing test or what is going on here?

lukahartwig (approved):

src/components/Label/Label.spec.tsx
snapsnapturtle: The test wasn't failing, because this property was just ignored (it's not valid css). I fixed it for completeness, so we actually test for the correct background color.

Issue #131, opened 2021-06-28
Crypto api (used by `nanoid`) is missing in some browsers


snapsnapturtle:
After deploying the Admin Panel with a new version of wave, we noticed an increase of a specific error `ReferenceError: 'crypto' is undefined`. This is very likely caused by the auto-generated ids for inputs (which are using the `nanoid` library). In the library, an error is thrown, when the crypto api does not exist in the browser. I have noticed that in IE 11 the crypto api does not exist.

-   `@freenow/wave` version: 1.4.0 (also applies in the latest version 1.6.0)

### Relevant code

```js

function Example = () => (
    <Input />
);
```

### What was expected to happen?

The component should render normally with a randomly generated id to link the label and the input component.

lukahartwig:
[There is a section on the nanoid README addressing this](https://github.com/ai/nanoid#ie). Should be easy to add this polyfill to the library. Correct me if I'm wrong but I think polyfills count as side effects so we might need to modify the `sideEffects` field in `package.json` as well.

PR #139, opened 2021-06-29
Add the crypto api polyfill for IE 11

snapsnapturtle:
**What:**
Adding a polyfill for IE11 and the Crypto api, closing #131.
​
**Why:**
After introducing the automatically generated ids for Input components using `nanoid`, we noticed a sharp increase of errors in our project. With some investigation, we found that nanoid requires the crypto api to be available on `window.crypto` (https://github.com/ai/nanoid#ie).
​
**How:**
Adding a specific `polyfill.ts` file to include this change to make IE 11 work with the design system again without throwing errors.

### DIFF
--- a/docs/browser-support.mdx
+++ b/docs/browser-support.mdx
@@ -7,8 +7,7 @@ cardSubHeadline: Creating a consistent experience across browsers and devices.
 
 ## Supported Devices
 
-The design system should be universally accessible, which means that the components should work in the browsers and devices used to access
-the project, which is using the design system. Components might not look perfect in all scenarios, but everything should still be
-accessible to the user.
+The design system should be universally accessible. Components might not look perfect in all scenarios, but everything
+should still be accessible to the user.
 
 **Currently, we are supporting the latest 2 versions of Chrome, Firefox, Safari, Edge and Internet Explorer 11.**
--- a/docs/browser-support.mdx
+++ b/docs/browser-support.mdx
@@ -1,0 +1,14 @@
+---
+name: Browser support
+route: /browser-support
+cardHeadline: Browser support
+cardSubHeadline: Creating a consistent experience across browsers and devices.
+---
+
+## Supported Devices
+
+The design system should be universally accessible, which means that the components should work in the browsers and devices used to access
+the project, which is using the design system. Components might not look perfect in all scenarios, but everything should still be
+accessible to the user.
+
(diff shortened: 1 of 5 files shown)


docs/browser-support.mdx
```diff
@@ -0,0 +1,14 @@
...
+name: Browser support
+route: /browser-support
+cardHeadline: Browser support
+cardSubHeadline: Creating a consistent experience across browsers and devices.
+---
+
+## Supported Devices
+
+The design system should be universally accessible, which means that the components should work in the browsers and devices used to access
+the project, which is using the design system. Components might not look perfect in all scenarios, but everything should still be
```
duranmla, line 11: Can we re-phrase this a bit? 
`which means that the components should work in the browsers and devices used to access
the project, which is using the design system` or it is just my English? haha

Perhaps, `The design system should be universally accessible. Components might not look perfect in all scenarios, but everything should still be accessible to the user` _(so we just drop that sentence)_

duranmla (approved):
🚀

package.json
```diff
@@ -5,7 +5,9 @@
     "main": "lib/cjs/index.js",
     "typings": "lib/types/index.d.ts",
     "module": "lib/esm/index.js",
-    "sideEffects": false,
+    "sideEffects": [
+        "src/polyfills.ts"
```
lukahartwig, line 9: I think this needs to point to the file in the `lib` folder, since that will be bundled. So in our case it should include both the `cjs` and `esm` version.

src/polyfills.ts
```diff
@@ -0,0 +1,5 @@
+// create an alias for the crypto api for IE11
```
lukahartwig: Maybe also mention the library that needs this, so we know that we can remove this in case we don't use `nanoid` anymore.

lukahartwig (changes requested):

src/polyfills.ts
snapsnapturtle: that makes sense, I've added it to the comment

package.json
snapsnapturtle: I wasn't sure what exactly this is pointing towards, but it makes sense to use the compiled files. Thanks

lukahartwig (dismissed):

docs/browser-support.mdx
snapsnapturtle: That's fine with me, I've updated the documentation!

duranmla (approved):

lukahartwig (approved):

PR #142, opened 2021-07-06
Update test fixture dependencies (webpack)

snapsnapturtle:
We have some dependencies in the test fixtures of this project, which are detected by Github. This does not have any impact on the compiled version or actual source, it's just related to our tests (no need to worry here).

**What:** Update dependencies for test fixtures (mainly upgrading from webpack 4 to version 5).
​
**Why:** So we can remove the dependabot alerts from this repository, which will allow us to detect if we encounter an actual security issue much easier.
​
**How:** Upgrading webpack and plugins for the test fixture projects

### DIFF
--- a/fixtures/babel-build/package-lock.json
+++ b/fixtures/babel-build/package-lock.json
@@ -1,8 +1,4473 @@
 {
     "name": "hello-world",
     "version": "0.0.0",
-    "lockfileVersion": 1,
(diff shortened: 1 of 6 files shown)


duranmla:
I don't grasp how dependabot can be an issue. At a first sight, it looks like a good idea to have it.

duranmla (approved):

PR #147, opened 2021-07-08
feat: remove the deprecated block property for buttons

### LINKED ISSUE #16: Remove deprecated "block" property from Button
After deprecating the "block" property and replacing it with the more generic "width" property, we now want to remove the old property for the next major release and update the documentation accordingly

snapsnapturtle:
This is the first of a few PRs to mark the next major version for wave. I have introduced a migration guide and a codemod framework, which we can use to enable developers to seamlessly migrate to the next major version.

It's using `jscodeshift` to transform the files, which is the de-facto standard when it comes to writing codemods for javascript-based projects. Tried it in three projects and found a few places that were still using the `block` property. 

Let me know what you think of this approach and if you're willing to invest the time of writing codemods for all breaking changes. From experience, having a migration guide is so helpful when upgrading to a new version, because you don't have to dig through all the changes to figure out all the changes you have to perform.

I will also have to target the `next` branch when this should continue.

Closes #16

### DIFF
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -29,7 +29,7 @@ jobs:
               run: npm run test
 
     release:
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/next'
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


duranmla (approved):
TIL `jscodeshift`. Thanks for the effort!

div-Leo (approved):

lukahartwig (approved):

Issue #149, opened 2021-07-09
How to implement form controls? Controlled/Uncontrolled components


duranmla:
### Support questions

There are some scenarios where folks need to have control over the value of a form control in order to be able to satisfy the requirements, currently, most of our implementations are uncontrolled components that encapsulate the state value within the implementation and can't be accessed from outside.

**Should we aim to have components that work out of the box but also expose a prop to control the state if needed?** 

_Similar to the approach described at https://egghead.io/lessons/react-make-controlled-react-components-with-control-props_

snapsnapturtle:
@duranmla Thanks for bringing this topic up as an issue. I think most of our components are controlled components (or at least support that api). I'm a big fan of controlled components as they seem to fit very well in the React landscape for components to always render deterministically. Let's find the components that don't match this and update them (where possible) so wave is more consistent overall

div-Leo:
@snapsnapturtle that is true, most of the components we use are controlled components. But what @duranmla was pointing out is a step further.  The idea is to have a component that works in isolation with it's own state that does not depend on its environment and also exposes props to control its state and relative renders. So far the ones we have don't work in isolation.
The video explains the idea behind it quite well.

Here you have an example of a possible implementation: https://github.com/freenowtech/wave/pull/153
Feel free to start the discussion there, so we can decide on a common strategy.

PR #154, opened 2021-07-13
Feat/search controlled component

div-Leo:
**What:**
Make Search component controlled with control props.
​
**Why:**
So far the Search component deals with the input value inside the component itself, it accepts a default value but then it's controlled by the internal state. Sometimes we need to update the value from the container and this is not possible with the current implementation. 
e.g. use case: I type a name in the search, then submit, now I can't empty the input 
​
**How:**
Exposes `value` and `setValue` props. Internally in the component, we check if we have a `propValue` and use it as the display value otherwise we use the one from the internal state. To update the value on the input `onChange` event we call either the props one (if we have it) or the internal one.

_**NB** This implementation is not a breaking change, it's completely compatible with previous versions_ 

**Update:**
This PR includes some changes to make controlled also the dropdown. 
The way we pick the state to change now is implemented using a hook `useControlledState`.

Here in the gif you can see how we clear and populate the input from the outer component, and also we close the dropdown when we pick our option. 
jBl4ZtWBYc


src/components/Search/Search.tsx
```diff
@@ -258,7 +273,7 @@ export const Search = ({
                         aria-label="clear-search"
                         style={{ margin: '1rem', marginLeft: 'auto', cursor: 'pointer', display: 'flex' }}
                         onClick={() => {
-                            setValue('');
+                            isOnControl ? setPropsValue?.('') : setStateValue('');
```
duranmla, line 276: Maybe. instead of doing this ternary on every step, we need to update the value we can have a `updateValue()` that allows us to have the change of the value in a single place.

```
const updateValue = (value) => {
    propsValue !== undefined ? setPropsValue?.('') : setStateValue('')
}
```

duranmla (approved):
For me this makes sense, and that was the reason for https://github.com/freenowtech/wave/issues/149 let's see what other folks say about it and make sure we align before moving forward so we have a clear principle on "how" to implement components

src/components/Search/Search.tsx
```diff
@@ -67,9 +67,13 @@ const StyledInput = styled(Input)`
...
     results?: React.ReactNode[];
     /**
-     * Sets the default value
+     * Sets the value
      */
     value?: string;
+    /**
+     * Function to set the value
+     */
+    setValue?: (value: string) => void;
```
lloydaf, line 76: IMHO `onInputChange` is a better alternative to `setValue`. Imo the `Search` should not care about what goes on around it. Whether we set value or call an API or whatever, should be left to the user (of the `Search`) to decide, so `onInputChange` seems more appropriate to me. You can use it the same way,
`<Search value={value} onInputChange={setValue} />`

lloydaf (changes requested):
IMHO this is a little more complicated than it needs to be. I think we can handle this with the existing props, I don't think we need to manage two states and have a flow of control logic between them. The problem I have with the current approach is that I am forced to manage it, and if I don't manage it, then the search won't work. I can't just provide a default and forget about it. Here's what I think can be a potential quick solution:
```jsx
    const [value, setValue] = React.useState<string>('');
    useEffect(() => setValue(propsValue),[propsValue]);
```

This way, we are updating the internal state value every time the external prop changes, hence making it *"controlled"*. Changes to the input text can be emitted using `onInputChange`, and the container can change `value` or store it in a different place or do whatever the user wants with it.

div-Leo:
@lloydaf, I understand your proposal, but I don't think is a common practice in React to have a useEffect to "emit" changes to the container. Usually, the method to follow is the deterministic approach (top => down), so we keep the state in the container (smart component) and pass info down in the tree to the presentational components (dumb). In this case, we need a little extra flexibility so the goal is to make both "smart components" but the container has the priority.
This is the reason why we choose between the parent props or the internal state.

lloydaf:
So in this case, `useEffect()` isn't really emitting to the container, it's just ensuring that the internal state of the `Search` stays in sync with the value that is in `props`. We're emitting when we change the input, using `onInputChange`, but that's similar to what the current proposal is doing anyway right? We're calling `setPropsValue`, which is actually an example of us emitting, right?

snapsnapturtle:
I did a bit of research on our existing components and how other design systems are handling this issue. Generally speaking: **The majority of recommendations say to provide controlled inputs**. Of course, the Search is a special case, but I would agree with this statement that our components should be controlled by default. Our Input component for example is also controlled.

Using a controlled component removes out the magic and prevents unintentional side effects from happening. I would argue, that in the case of the `Search` component, we should also follow this approach and allow users to control the state, regardless of the internal state. In hindsight, the `value` prop should probably be named `defaultValue` to indicate it's true purpose.

If we decide to make this a controlled component and remove the "defaultValue", it would become a breaking change. We already have others lined up, so we could relatively soon also release them together as a v2. I have a proposal up for breaking changes in #147

src/components/index.ts
```diff
@@ -45,3 +45,4 @@ export { Search, SearchProps } from './Search/Search';
 export { Password } from './Password/Password';
 export { VisuallyHidden } from './VisuallyHidden/VisuallyHidden';
 export { RadioButtonProps } from './RadioButton/RadioButtonProps';
+export { Offcanvas, OffcanvasProps} from './Offcanvas/Offcanvas';
```
snapsnapturtle, line 48: Just noticed this, but there's a space missing at the end of the export which should have been added by prettier. Can you check that it's actually working or if we have an issue?

div-Leo: It does work in general, because I saw it did format some part of the code following our prettier. 
I'm not sure why it didn't in that case.

div-Leo:
@snapsnapturtle this version is retroactive, it works just as fine as before, plus it makes it controlled.

src/components/Search/Search.tsx
snapsnapturtle: I agree with @lloydaf here, `onInputChange` is more in line with other components of the design system and will not add confusion to this component

snapsnapturtle (approved):

duranmla (approved):

src/components/index.ts
```diff
@@ -45,3 +45,4 @@ export { Search, SearchProps } from './Search/Search';
 export { Password } from './Password/Password';
 export { VisuallyHidden } from './VisuallyHidden/VisuallyHidden';
 export { RadioButtonProps } from './RadioButton/RadioButtonProps';
+export { Offcanvas, OffcanvasProps } from './Offcanvas/Offcanvas';
```
duranmla, line 48: Not sure why we have this tho.

lloydaf (approved):

PR #158, opened 2021-07-19
Feat/add icon minus

div-Leo:
**What:**
Add minus icon
​
**Why:**
The minus icon is missing and it's required in one of the layouts
​
**How:**
Added the minus .svg and used the `generate` script.
​
**Media:**
Screenshot 2021-07-19 at 12 05 58


assets/icons/basic/minus.svg
```diff
@@ -0,0 +1,13 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
+    <g fill="none" fill-rule="evenodd">
+        <g fill="#637689" fill-rule="nonzero">
+            <g>
```
snapsnapturtle, line 4: Just a small style thing, but maybe you can remove the empty g tags?

src/components/index.ts
```diff
@@ -45,3 +45,4 @@ export { Search, SearchProps } from './Search/Search';
 export { Password } from './Password/Password';
 export { VisuallyHidden } from './VisuallyHidden/VisuallyHidden';
 export { RadioButtonProps } from './RadioButton/RadioButtonProps';
+export { Offcanvas, OffcanvasProps} from './Offcanvas/Offcanvas';
```
snapsnapturtle, line 48: I'm not sure where this is from now, but there's a space missing before the closing `}`. Wasn't this in another PR already?

snapsnapturtle (approved):

src/components/index.ts
div-Leo: I think I have the problem that I can't pull from the wave remote so it's like my remote/branch thinks to be ahead then the wave one. 

I'm not sure why... I just fixed it here hope is not going to create conflicts in the other.

lloydaf (approved):

snapsnapturtle:
@div-Leo there seems to be a merge conflict with the current main branch. Do you think you can resolve it? I'll merge it right after, thanks!

rafael-sepeda (approved):
We have this icon in the design library already, so it would be great to have it in Wave!

PR #163, opened 2021-07-26
feat: replace tslint with eslint

### LINKED ISSUE #4: Migrate from tslint to eslint
Since tslint has been deprecated, we should switch to the successor (eslint). This requires us to upgrade the build steps and replicate the same settings we currently have in tslint, including the auto fixers.

snapsnapturtle:
Well, this is going to be a hell to review. Sorry in advance. For some time, the issue to migrate to eslint has been laying in our backlog and I thought it would be better to do it now, rather than add more source code that needs to be updated. So here it is: ~160 files changed and a lot of things to review. I learned a lot about types on the way and even found some bugs that we weren't aware of. Closes #4 

There are a couple of things that need some explanation and we should evaluate if this is the right way forward. For the beginning I started off with one of the most used linter configurations and added our special rules to it.

We should do some intensive testing before actually merging this PR as I had to update a few pieces of business logic on the way.


docs/components/ImageModal.tsx
```diff
@@ -11,7 +17,7 @@ const ClickableImage = styled.img`
     }
 `;
 
-export const ImageModal = ({ src, alt }) => {
+export const ImageModal: FC<ImageModalProps> = ({ src, alt }: ImageModalProps) => {
```
christopherrolfe198, line 20: From what I've read and seen from using the library, fC causes a lot of typing issues and it seems that the community has had some discussion around dropping it. This might not be the best place to have this discussion but it might be an idea to stop using FC generally and add our own children prop types.

snapsnapturtle: Agreed, I've read about the downsides as well. I was thinking about replacing them all here as well, but this would have made the PR less about getting it compliant with the linter and more about improving it as well. It's already huge as it is. 

There are some really good alternative interfaces you can use when you want to use children in the react api as well :)

martimalek: I've also read about the downsides but I'm not clear on what the best alternative is, could you share what is your approach?

christopherrolfe198: So the alternative is to type the children yourself explicitly, make it clear what the component expects, if you need to match the current behaviour of FC you can type it with React.ReactNode (if I remember correctly).

I think React does now provide some better built in types as well but I'm less familiar with them. I'm going to resolve this thread based on what we've discussed, I agree that this isn't the place to fix this but just to aware that we might cause additional ts errors for end users with it.

martimalek (approved):

src/components/Datepicker/DatepickerSingleInput.spec.tsx
```diff
@@ -64,12 +64,12 @@ describe('DatepickerSingleInput', () => {
         const formattedDate = format(intendedDate, 'dd/MM/yy');
 
         // Simulate an input change
-        await userEvent.type(input, formattedDate);
+        userEvent.type(input, formattedDate);
 
         expect(mockErrorTextHandler).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
snapsnapturtle: Thanks for bringing this up. In this case, I just removed the await from the userEvent, because the function isn't returning a promise that can be "awaited". I'd love to follow up with a PR to improve the tests as well. It looks like this huge change is a good opportunity to evaluate the existing code base.

docs/components/ImageModal.tsx
snapsnapturtle: You can check out the typings for react (yes they are not that great to read) for how it's done internally. From my experience and reading, I'd suggest just making the function return a `ReactElement | null`, which is in line with what the function from [`FC`/`FunctionalComponent`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L544) returns.

src/components/Datepicker/DatepickerSingleInput.spec.tsx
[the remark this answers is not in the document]
martimalek: I've created #173 and will mark this as resolved

PR #170, opened 2021-08-26
Improves Text component

### LINKED ISSUE #103: Document intended usage for disabled Text
Relates to #96

nlopin:
**What:**
* Adds new `secondary` props and deprecates the `weak`
* Migrates `Text` components to color tokens
* Improves the documentation
​

Closes #103


src/components/Text/Text.tsx
```diff
@@ -33,28 +33,33 @@ interface TextProps
...
      * Adjust color to display a disabled text element
      */
     disabled?: boolean;
 }
 
-function determineTextColor({ weak, inverted, disabled }: TextProps) {
+function determineTextColor({ weak, secondary, inverted, disabled }: TextProps) {
     if (disabled) {
-        return inverted ? Colors.AUTHENTIC_BLUE_550 : Colors.AUTHENTIC_BLUE_350;
+        return inverted ? SemanticColors.text.disabledInverted : SemanticColors.text.disabled;
```
snapsnapturtle, line 51: Please use the theme get method, so we have a single place to update the colors (with the theme). In addition this makes it possible for users to consistently override a single color across the components (see here https://github.com/freenowtech/wave/blob/main/src/components/Button/Button.tsx#L23).

```suggestion
        return inverted ? get('semanticColors.text.disabledInverted')(props) : get('semanticColors.text.disabled')(props);
```

Be sure to also pass all the props to the function, instead of descructuring them in L49. This way, the required props (theme for example) can be passed correctly to the get function.

src/components/Text/docs/Text.mdx
```diff
@@ -20,6 +20,11 @@ The Text component is a wrapper component that will apply typography styles to t
 
 <StyledSystemLinks component="Text" supportedProps={ ['margin', 'fontSize', 'fontWeight', 'fontFamily', 'textAlign'] }/>
 
+## Properties
+
+<TextPropsTable />
```
snapsnapturtle, line 25: Nice, thanks for making it more consistent 🎊

src/components/Text/Text.tsx
```diff
@@ -33,28 +33,33 @@ interface TextProps
     inverted?: boolean;
     /**
      * Adjust color to indicate secondary information
+     * @deprecated
      */
     weak?: boolean;
+    /**
+     * Adjust color to indicate secondary information
+     */
+    secondary?: boolean;
```
snapsnapturtle, line 42: Question: How can we ensure the user knows that the `disabled` prop overrides the `weak` and `secondary` props? Maybe we can just add this to the documentation and the props table, so that the users are aware of it and don't have to look it up in the source code.

src/components/Text/docs/TextPropsTable.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { PropsTable } from '../../../docs/PropsTable';
+
+export const TextPropsTable = () => {
+    const props = [
```
snapsnapturtle, line 5: Should we for now include the `weak` prop in the props table? I don't have a strong opinion, but it would be consistent and allow users to find the things they're currently already using.

src/components/Text/Text.tsx
```diff
@@ -33,28 +33,33 @@ interface TextProps
...
-function determineTextColor({ weak, inverted, disabled }: TextProps) {
+function determineTextColor({ weak, secondary, inverted, disabled }: TextProps) {
     if (disabled) {
-        return inverted ? Colors.AUTHENTIC_BLUE_550 : Colors.AUTHENTIC_BLUE_350;
+        return inverted ? SemanticColors.text.disabledInverted : SemanticColors.text.disabled;
     }
 
-    if (weak) {
-        return inverted ? Colors.AUTHENTIC_BLUE_350 : Colors.AUTHENTIC_BLUE_550;
+    if (secondary || weak) {
```
snapsnapturtle, line 54: We have an amazing helper function that sends a console message in the dev environment to let the developers know about the deprecation: https://github.com/freenowtech/wave/blob/main/src/utils/deprecatedProperty.ts. I would suggest we use it here as well, when the `weak` prop is set. It will make upgrading easier and encourage teams to use the new `secondary` prop instead.

src/components/Text/Text.tsx
```diff
@@ -33,28 +33,33 @@ interface TextProps
     inverted?: boolean;
     /**
      * Adjust color to indicate secondary information
+     * @deprecated
```
snapsnapturtle, line 36: Can we add a comment, explaining what prop to use instead? Otherwise the users will assume that there is no replacement for this property.

nlopin: Done

src/components/Text/Text.tsx
nlopin: Thanks for reminding me about that function 👍

src/components/Text/docs/TextPropsTable.tsx
nlopin: I added the prop to the table with the deprecated label

src/components/Text/Text.tsx
nlopin: I did two things:

- raised the `disabled` prop higher in the props table to implicitly show the precedence
- added an explicit text to the description

nlopin:
I'm sorry it took that much time. All the comments are addressed, please, take a look

rafael-sepeda:
@nlopin I left some comments

rafael-sepeda (changes requested):

src/components/Text/Text.spec.tsx
```diff
@@ -30,10 +30,24 @@ describe('Text', () => {
...
+                color: ${Colors.AUTHENTIC_BLUE_350};
+            `);
+        });
+    });
+
     describe('if the disabled property is set', () => {
         it('has the disabled color', () => {
             expect(render(<Text disabled />).container.firstChild).toHaveStyle(`
-                color: ${Colors.AUTHENTIC_BLUE_350};
+                color: ${Colors.AUTHENTIC_BLUE_200};
```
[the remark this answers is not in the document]
nlopin: For me, the test should check the final value assigned to avoid occasional changes in the semantic colors assignment. 

But I don't have a strong opinion here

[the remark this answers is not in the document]
rafael-sepeda: @nlopin the thing is that with semantic colors they should be the primary source of truth, rather then the brand colors (`Colors.AUTHENTIC_BLUE_350`). Because the brand colors might change, but the semantic should stay, because it reflects the purpose of the element (text, in this case). In FREE NOW there was, already, several color rebranding and, by the way, soon will be another round. So in best we should link all the things to semantic colors, and then, if we need to do that, we can change theirs values in one single place.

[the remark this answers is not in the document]
nlopin: Makes sense, I'll change the test to use the semantic colors

nlopin:
@rafael-sepeda take a look

rafael-sepeda (approved):
Nice !

martimalek (approved):
Looks great!

christopherrolfe198 (approved):
I'm approving as this keeps showing up for me and I can't seem to remove myself from the reviewers xD
Others have approved so it should be fine.

jonotrujillo (approved):

PR #171, opened 2021-08-31
Feature/add indeterminate state

lloydaf:
**What:**

This PR aims to add the `indeterminate` state to the wave `Checkbox` component
​
**Why:**

​We have a usecase where we have a table and a checkbox in the header, which should be `selected` only if all items in the table are selected, `unselected` only if all items in the table are unselected, and `indeterminate` otherwise
**How:**

​1. The [spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate) provides an `indeterminate` attribute to be set using JS
2. Introduce a new prop called `indeterminate` to the `Checkbox` component
3. Update the styles in `Checkmark` and `LabelWrapper` to reflect this
4. Update the tests and the documentation to reflect this
**Media:**

Watch me loom [here](https://www.loom.com/share/6e915eb004564d45b29a5f4973a39be6)
**Checklist:**

- [x] Ready to be merged

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -1,4 +1,4 @@
-import React, { DetailedHTMLProps, FC, InputHTMLAttributes, ReactNode } from 'react';
+import React, { DetailedHTMLProps, FC, InputHTMLAttributes, ReactNode, useEffect } from 'react';
 import styled from 'styled-components';
 import { MarginProps, ResponsiveValue } from 'styled-system';
 
@@ -29,6 +29,10 @@ interface CheckboxProps
      * Define size of the checkbox component, defaults to large
      */
     size?: ResponsiveValue<'small' | 'medium' | 'large'>;
+    /**
+     * Decides whether the checkbox is in an "indeterminate"(partially selected) state
+     */
+    indeterminate?: boolean;
 }
 
 const WithTapAreaWrapper = styled.div<Pick<CheckboxProps, 'textVerticalAlign'>>`
@@ -45,7 +49,7 @@ const Checkbox: FC<CheckboxProps> = props => {
     const { classNameProps, restProps: withoutClassName } = extractClassNameProps(props);
     const { marginProps, restProps } = extractWrapperMarginProps(withoutClassName);
 
-    const { disabled, error, label, textVerticalAlign, size, ...rest } = restProps;
+    const { disabled, error, label, textVerticalAlign, size, indeterminate, ...rest } = restProps;
     let dynamicLabel: ReactNode = label;
 
     if (typeof label === 'string') {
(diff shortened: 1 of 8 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -56,8 +60,15 @@ const Checkbox: FC<CheckboxProps> = props => {
         );
     }
 
+    const checkboxRef = React.createRef<HTMLInputElement>();
+
+    useEffect(() => {
```
snapsnapturtle, line 65: What is the useEffect in combination with the ref accomplishing here? I would assume that you would be able to pass the prop directly to the checkbox

snapsnapturtle:
Thanks for contributing!

nlopin:
I was thinking that it would be great to have this as a new component. The use case looks specific enough to have such complexity outside of the plain `Checkbox` component. Such approach also reduces the possibility of bugs for current users of the `Checkbox`.

src/components/Checkbox/Checkbox.tsx
lloydaf: Currently, browsers don't support setting this through the HTML, and must be set though javascript
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate

lloydaf:
> I was thinking that it would be great to have this as a new component. The use case looks specific enough to have such complexity outside of the plain `Checkbox` component. Such approach also reduces the possibility of bugs for current users of the `Checkbox`.

@nlopin personally I think that's a bad idea, this is an attribute that's natively supported by the spec for elements of the type `<input type="checkbox">`, so imho it should exist in our `Checkbox` component. If this was something that's not part of the spec I might have agreed with you (as a special use case probably demands another component), but as far as I understand this is a very common use case for checkboxes.

snapsnapturtle:
> > I was thinking that it would be great to have this as a new component. The use case looks specific enough to have such complexity outside of the plain `Checkbox` component. Such approach also reduces the possibility of bugs for current users of the `Checkbox`.
> 
> @nlopin personally I think that's a bad idea, this is an attribute that's natively supported by the spec for elements of the type `<input type="checkbox">`, so imho it should exist in our `Checkbox` component. If this was something that's not part of the spec I might have agreed with you (as a special use case probably demands another component), but as far as I understand this is a very common use case for checkboxes.
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate

I agree. Since it's natively supported, it makes sense to also support it in our "native" component 👍

src/components/Checkbox/Checkbox.tsx
snapsnapturtle: Thanks for the explanation. Can you add this comment to the code as well and explain why it is the way it is? That way future devs aren't confused

lloydaf: Done!

snapsnapturtle (approved):
LGTM

christopherrolfe198 (approved):

nlopin (approved):
I think we need OK from one designer

rafael-sepeda (approved):

PR #172, opened 2021-09-08
feat: added Accordion component

pgr1:
What:
Make Accordion component.
​
Why:
After checking that different products like DMT,  Campaigns and Quests use it we decide to add it to wave

How:
Create new component and documentation 

Screenshot 2021-09-08 at 09 57 00

Screenshot 2021-09-08 at 09 57 50


src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+const RenderedSection = styled(Box)`
+    :last-child ${HorizontalDividerBottom} {
+        display: inherit;
+    }
+`;
+
+const Accordion = ({ heading, description, info, buttonLabel, variant, expanded, children }: AccordionProps) => (
+    <RenderedSection role="group">
+        <HorizontalDividerTop />
+        {variant === 'compact' ? (
```
martimalek, line 39: I'd use an enum for the variants so that we don't need to write the constants manually each time, the goal would be to  just do `variant === AccordionVariants.Compact`

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <Compact label={heading ? heading : ''} description={description} expanded={expanded}>
```
martimalek, line 40: `heading ? heading : ''` can be simplified to `heading || ''` or we could just use default props

src/components/Accordion/components/ChevronDown.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
+import { ChevronDownIcon } from '../../../icons';
+import { Colors } from '../../../essentials';
+
+export const ChevronDown = styled(ChevronDownIcon)`
+    color: ${props => (props.color ? props.color : Colors.AUTHENTIC_BLUE_900)};
```
martimalek, line 6: `props.color ? props.color : Colors.AUTHENTIC_BLUE_900)` can be simplified to `props.color || Colors.AUTHENTIC_BLUE_900)`

src/components/Accordion/components/ChevronUp.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
+import { ChevronUpIcon } from '../../../icons';
+import { Colors } from '../../../essentials';
+
+export const ChevronUp = styled(ChevronUpIcon)`
+    color: ${props => (props.color ? props.color : Colors.AUTHENTIC_BLUE_900)};
```
martimalek, line 6: Same

src/components/Accordion/docs/Accordion.mdx
```diff
@@ -0,0 +1,50 @@
...
+route: /components/accordion
+---
+
+import { Playground } from 'docz';
+import { AccordionPropsTable } from './AccordionPropsTable.tsx'
+import { Accordion } from '../Accordion.tsx';
+import { Box } from '../../Box/Box.tsx';
+
+# Accordion
+Accordion component is use to display large amount of content in a compress and progressive way.
```
martimalek, line 13: Typo, maybe we could rephrase it to something like:

> The Accordion component is used to display a large amount of content in a compressed and progressive way.

src/components/Accordion/docs/AccordionPropsTable.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+import { PropsTable } from '../../../docs/PropsTable';
+
+export const AccordionPropsTable = () => {
+    const props = [
```
martimalek, line 5: If this is a static array it could be defined outside to not recreate the array on each call

martimalek (changes requested):

src/components/Accordion/Accordion.tsx
pgr1: Got your point but this is the only part `variant` is validated, so I prefer to leve as a string

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+    expanded?: boolean;
+    children: ReactNode;
+}
+
+const HorizontalDivider = styled(Box)`
+    border: 0;
+    border-top: solid 1px ${Colors.AUTHENTIC_BLUE_200};
+`;
+
+const HorizontalDividerTop = HorizontalDivider;
```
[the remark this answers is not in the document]
pgr1: I need two different dividers, the top divider is always rendered and the bottom divider is only rendered in the last position.

rafael-sepeda (approved):
Big thanks for pushing this topic!

martimalek (approved):

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { ReactNode } from 'react';
+import styled from 'styled-components';
+
+import { Colors } from '../../essentials';
+import { Box } from '../Box/Box';
+import Compact from './components/Compact';
+import DefaultPanel from './components/Default';
+
+interface AccordionProps {
+    heading?: string;
```
christopherrolfe198, line 10: Are these types correct? Should we be limiting to a string or is it acceptable to pass through a react element here as well? Perhaps React.ReactNode fits better.

src/components/Accordion/components/Default.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { Box } from '../../Box/Box';
+import { Headline } from '../../Headline/Headline';
+import { Header } from './Header';
+import { ChevronUp } from './ChevronUp';
+import { ChevronDown } from './ChevronDown';
+import { Description } from './Description';
+import { PanelButton } from './PanelButton';
+
+interface Props {
+    label: string;
```
christopherrolfe198, line 15: Same as before, are these types correct?

src/components/Accordion/components/Default.tsx
```diff
@@ -0,0 +1,102 @@
...
+    border: solid 1px ${Colors.AUTHENTIC_BLUE_200};
+    border-radius: 5px;
+`;
+
+const DefaultPanel = ({ label, description, info, buttonLabel, expanded = false, children }: Props) => {
+    const [isOpen, setIsOpen] = useState<boolean>(expanded);
+
+    return (
+        <>
+            {isOpen ? (
```
christopherrolfe198, line 68: NIT: I personally prefer to not use ternary operators with react as it can be unclear why you are rendering something.

E.G. `{isOpen && <Component />}` and `{!isOpen && <Component />}` is clearer for me but I don't know if there are any conventions around this in the codebase @nlopin

christopherrolfe198 (changes requested):

src/components/Accordion/Accordion.tsx
pgr1: The `heading` is just the header title. I'll renamed it to avoid confusions

christopherrolfe198: Yeah but it doesn't make sense to me to restrict it, for example, if you wanted to include an icon within the heading or something along those types of lines would that be an issue?

src/components/Accordion/components/Default.tsx
pgr1: To be honest for me is more readable with a ternary. But both approaches are equivalent

src/components/Accordion/Accordion.tsx
pgr1: In the design the title only can be a text, that is why that case is not contemplated

martimalek: In the syncs with design we came to the conclusion that this component would be restricted in multiple aspects, e.g. title and label not accepting nodes, just strings. I suppose if in the future there appear new needs we would extend the component

christopherrolfe198: I'm a little confused as to why we'd restrict this now? I mean if a designer presents a design where we need something other than a string it creates more work later. Whereas having nodes now doesn't stop a string being entered. Also any use of this component will be checked by a designer.

martimalek: This was restricted by the designers, not by us. As far as i remember it was their initiative to follow this same pattern (with just a string) and commit to only using it in this manner, the idea in the early conversations about the Accordion was to accept a node as the header but it was changed. Of course in case a designer needs something other than a string then we'll need to change it but in principle it shouldn't happen. Please correct me @rafael-sepeda

rafael-sepeda: Sorry @martimalek for late reaction.

@christopherrolfe198 yes, for now we want to have want to have a limited set of variants. Agree that in general we might need more flexibility in the future. However to make the setup more easy to implement we decided to compromise the flexibility.

By design we have a limited list of variants
https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?node-id=1973%3A67543

src/components/Accordion/components/Compact.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    {isOpen && <Description mt="1" description={description} />}
+                </Box>
+                {!isOpen && <Description description={description} />}
+                <PanelButton isOpen={isOpen} />
+            </PanelHeader>
+            {isOpen && <Box mb="5">{children}</Box>}
+        </>
+    );
+};
+export default Compact;
```
snapsnapturtle, line 54: We usually don't have default exports in this project. There are a few reasons why named exports are preferred, mainly for maintainability down the road.

- [ ] Please use a named export for this component

src/components/Accordion/components/Compact.tsx
```diff
@@ -0,0 +1,54 @@
...
+    &:hover ${ChevronDown} {
+        color: ${Colors.ACTION_BLUE_1000};
+    }
+
+    &:hover ${ChevronUp} {
+        color: ${Colors.ACTION_BLUE_1000};
+    }
+`;
+
+const Compact = ({ title, description, expanded = false, children }: Props) => {
```
snapsnapturtle, line 35: To include the official children, you can use the `PropsWithChildren<T>` from the react library.
```suggestion
const Compact = ({ title, description, expanded = false, children }: PropsWithChildren<Props>) => {
```

src/components/Accordion/components/Default.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { Text } from '../../Text/Text';
+import { Box } from '../../Box/Box';
+import { Headline } from '../../Headline/Headline';
+import { Header } from './Header';
+import { ChevronUp } from './ChevronUp';
+import { ChevronDown } from './ChevronDown';
+import { Description } from './Description';
+import { PanelButton } from './PanelButton';
+
+interface Props {
```
snapsnapturtle, line 14: For re-usability you can use the `Pick` typescript interface, so you don't have to repeat the same props from the `AccordionProps`.

src/components/Accordion/components/Default.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    </Box>
+                    <PanelButton isOpen={isOpen} color={Colors.ACTION_BLUE_900}>
+                        <ButtonLabel>{buttonLabel}</ButtonLabel>
+                    </PanelButton>
+                </PanelHeader>
+            )}
+        </>
+    );
+};
+export default DefaultPanel;
```
snapsnapturtle, line 102: Same as for the `Compact` component

- [ ] Please use a named export for the `DefaultPanel`

src/components/Accordion/components/Header.tsx
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components';
+
+import { Colors } from '../../../essentials';
+import { Box } from '../../Box/Box';
+
+export const Header = styled(Box).attrs({ p: '2', color: Colors.AUTHENTIC_BLUE_900 })`
```
snapsnapturtle, line 6: _No change necessary at the moment (but welcome)_: In an effort to make use of the new semantic colors we introduced, I wanted to let you know about another way to use this variable in the code. It's what we want to encourage moving forward: Semantic/Functional colors. Instead of accessing the `Colors` variable, you will use the `get` function in the `themeGet.ts` file in this project to get a semantic color from our library. See reference here: https://wave.free-now.com/essentials/colors#functional-reference.

```
color: get('semanticColors.text.primary')
```

src/components/Accordion/docs/AccordionPropsTable.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+import { PropsTable } from '../../../docs/PropsTable';
+
+const props = [
+    {
+        name: 'heading',
```
snapsnapturtle, line 6: Please update the props to match the actual API of the component, so the documentation is up-to-date!

- [ ] Update the documentation component to reflect actual API

src/components/Accordion/Accordion.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+    it('render default variant closed', () => {
+        render(
+            <Accordion heading="Some heading" info="some info" description="some description" buttonLabel="button">
+                <p>Place holder</p>
+            </Accordion>
+        );
+        expect(screen.getByText('Some heading')).toBeInTheDocument();
+        expect(screen.getByText('some info')).toBeInTheDocument();
+        expect(screen.getByText('some description')).toBeInTheDocument();
+        expect(screen.getByText('button')).toBeInTheDocument();
```
snapsnapturtle, line 27: Suggestion: We could add one more assertion that ensures that the `Place holder` is also NOT in the document.

snapsnapturtle (commented):
Thank you for contributing this component and also adding a test 🎉

src/components/Accordion/Accordion.tsx
martimalek: I think that because @christopherrolfe198 is on vacation and design has approved it we can mark this as resolved

src/components/Accordion/components/Default.tsx
martimalek: Same,we can mark it as resolved

src/components/Accordion/components/Default.tsx
martimalek: I would personally also go for the ternary, we could mark this as resolved

src/components/Accordion/components/Compact.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const [isOpen, setIsOpen] = useState<boolean>(expanded);
```
garthtee, line 36: Can we add an effect here to update the isOpen state, in the case that expanded changes on another render? 

```suggestion
    const [isOpen, setIsOpen] = useState<boolean>(expanded);
    
    React.useEffect(() => {
        if (isOpen !== expanded) setIsOpen(expanded);
    }, [isOpen, expanded]);
```

src/components/Accordion/components/Default.tsx
```diff
@@ -0,0 +1,102 @@
...
+    }
+`;
+
+const PanelBody = styled(Box).attrs({ my: '3' })`
+    border: solid 1px ${Colors.AUTHENTIC_BLUE_200};
+    border-radius: 5px;
+`;
+
+const DefaultPanel = ({ title, description, info, buttonLabel, expanded = false, children }: Props) => {
+    const [isOpen, setIsOpen] = useState<boolean>(expanded);
```
garthtee, line 64: Same here with the useEffect when expanded changes

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+    display: none;
+`;
+
+const RenderedSection = styled(Box)`
+    :last-child ${HorizontalDividerBottom} {
+        display: inherit;
+    }
+`;
+
+const Accordion = ({ title, description, info, buttonLabel, variant, expanded, children }: AccordionProps) => (
```
garthtee, line 36: One of our requirements on this Accordion is to change the text when it's either opened or closed. Maybe we could let the parent component take care of the state in this case?

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <RenderedSection role="group">
+        <HorizontalDividerTop />
```
martimalek, line 38: I believe this divider should not be rendered when the Accordion is expanded

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+                title={title || ''}
+                description={description}
+                buttonLabel={buttonLabel}
+                info={info}
+                expanded={expanded}
+            >
+                {children}
+            </DefaultPanel>
+        )}
+        <HorizontalDividerBottom />
```
martimalek, line 54: I believe this divider should not be rendered when the Accordion is expanded

src/components/Accordion/Accordion.tsx
martimalek, line 38: I believe this divider should not be rendered when the Accordion is expanded

martimalek (changes requested):
There are a couple of new issues I've found

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+import Compact from './components/Compact';
+import DefaultPanel from './components/Default';
+
+interface AccordionProps {
+    title?: string;
+    description?: string;
+    info?: string;
+    buttonLabel?: string;
+    variant?: 'compact' | 'default';
+    expanded?: boolean;
```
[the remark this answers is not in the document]
pgr1: Done

src/components/Accordion/components/Compact.tsx
pgr1: Done

src/components/Accordion/components/Compact.tsx
pgr1: Done

src/components/Accordion/components/Default.tsx
pgr1: Done

src/components/Accordion/docs/AccordionPropsTable.tsx
pgr1: Done

src/components/Accordion/Accordion.spec.tsx
pgr1: Done

src/components/Accordion/components/Default.tsx
pgr1: Done

src/components/Accordion/components/Compact.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+interface Props {
+    title: string;
+    description?: string;
+    expanded?: boolean;
+    children?: ReactNode;
+}
+
+const PanelHeader = styled(Header).attrs({ px: '0' })`
+    // @ts-ignore
```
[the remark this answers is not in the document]
pgr1: Added a comment wit te error.
Apparently ts don't get the Headline types inside style component

snapsnapturtle (approved):

PR #178, opened 2021-09-30
docs: add versioning policy

snapsnapturtle:
**What:**
In this PR we are documenting our versioning policy and add details on what is considered a major, minor and patch versions. 
​
**Why:**
With this additional documentation, we aim to clearly explain which change causes which version number to increase.


rafael-sepeda (approved):
Cool! I use this already in Figma, but not yet fully detailed

PR #179, opened 2021-10-08
improvement: select list component is better with react windowed select

### LINKED ISSUE #57: SelectList becoming really slow with large set of options
Hello, we've recently noticed that the SelectList component becomes really slow when populated with a large amount of options (multiple thousands).
Do you think it is worth introducing [react-window](https://github.com/bvaughn/react-window) in case a certain threshold of options is hit or rather think of different solutions than a Select in these cases?
I've played around a bit trying to overwrite the MenuList, similar to what [react-windowed-select](https://github.com/jacobworrel/react-windowed-select) does, but getting it correctly including styling a stuff seems not very trivial.

https://user-images.githubusercontent.com/9133431/116364428-79b25d00-a804-11eb-84fa-dd0911007259.mov

In the video you can see how slow it opens/closes and how it is very clunky when scrolling and hovering.

lloydaf:
**What:**

This PR aims to improve the performance of the `SelectList` component
​
**Why:**

Closes #57 
​
**How:**

Uses [react-windowed-select](https://github.com/jacobworrel/react-windowed-select) to not render all the elements at once
​
**Media:**

Watch me loom [here](https://www.loom.com/share/d85757d25c5d432687f893f7c705a6a5)

**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23457,9 +23457,9 @@
             }
         },
         "memoize-one": {
-            "version": "5.1.1",
-            "resolved": "https://registry.npmjs.org/memoize-one/-/memoize-one-5.1.1.tgz",
-            "integrity": "sha512-HKeeBpWvqiVJD57ZUAsJNm71eHTykffzcLZVYWiVfQeI1rJtuEaS7hQiEpWfVVk18donPwJEcFKIkCmPJNOhHA=="
+            "version": "5.2.1",
+            "resolved": "https://registry.npmjs.org/memoize-one/-/memoize-one-5.2.1.tgz",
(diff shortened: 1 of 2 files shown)


lloydaf:
> Love the work here and I think that we should merge this as an improvement for now but I have a concern about how we don't control the version of react-select anymore. Like it'd be nicer if react-select was a peer dependency of react-windowed-select so that we could keep control of that. (because it might contain other performance benefits in the future that we'd want to take advantage of)

@christopherrolfe198 completely agree, however I think we have for ourselves a little time, the library seems to support `v4.x` as well of `react-select`. But I agree that it's not a permanent solution, maybe we can raise a request to maintain it as a peer dep 🤷🏻

PR #196, opened 2021-12-17
Support responsive `size` values for Headline

### LINKED ISSUE #189: Headline: make `size` a responsive property
In the design, we often have different font sizes defined for mobile and desktop versions of the UI. At the moment, there is no easy way to make it without explicit checks in the component. It would be great if we can provide a list or an object of sizes based on breakpoints to the `size` prop of the `Headline` component in a similar fashion we do with `p` and `m` props:

```jsx
<Headline as='h2' size={['m', 'l']}>Text</Headline>
<Headline as='h2' size={{_: 'l', small: 'm'}}>Text</Headline>
```

nlopin:
**What:**

Change the size of the headline depending on the viewport size. 

You can pass a responsive value to the `size` prop of the `Headline` component and it will change the size depending on the viewport.  For example, `<Headline size={{_: 's', m: 'xl'}}`>headline</Headline>`.
​
**Why:**

Closes #189
​

**How:**

* Used `system` function from the `styled-system` to define the list of possible values for headline CSS properties `line-height` and `font-size`
* Pass the `size` prop to the style function returned by `system` to get responsive styles
​

**Media:**

https://www.loom.com/share/62e01f42c7d1449e9adbf1a7eb227db6

**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/src/components/Headline/docs/Headline.mdx
+++ b/src/components/Headline/docs/Headline.mdx
@@ -28,7 +28,6 @@ the semantics while following the design.
 
 <HeadlinePropsTable />
 
-
 ## Default headlines
 
 <ItemWrapper>
@@ -60,9 +59,24 @@ the semantics while following the design.
     </Headline>
 </ItemWrapper>
 
-## Examples
+## Responsive size
+
+The `size` property supports [responsive values](https://styled-system.com/responsive-styles/). Pass an array or an object
+and the design system use corresponding values for different viewport sizes.
+
+For example, the following snippet uses `s` size for mobile and `xl` size bigger screens:
+
+```jsx
+
+<Headline as="h2" size={{_: 's', medium: 'xl'}}>Small on mobile, big on bigger screens</Headline>
+
+```
+
+## Playground
+
 
 <Playground>
     <Headline as="h3">default headline h3</Headline>
     <Headline as="h1" size="xs">The smallest h1</Headline>
+    <Headline as="h4" size={{_: 'xs', medium: 'm'}}>Responsive headline</Headline>
 </Playground>
--- a/src/components/Headline/Headline.tsx
+++ b/src/components/Headline/Headline.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { ComponentPropsWithoutRef } from 'react';
-import styled, { css } from 'styled-components';
-import { compose, margin, MarginProps, textAlign, TextAlignProps } from 'styled-system';
+import styled from 'styled-components';
+import { compose, margin, system, MarginProps, textAlign, TextAlignProps, ResponsiveValue } from 'styled-system';
 import { Colors } from '../../essentials';
 import { theme } from '../../essentials/theme';
 import { get } from '../../utils/themeGet';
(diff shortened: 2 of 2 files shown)


rafael-sepeda:
@nlopin I have a proposal: since the behavior of the headline, in this case, will depend on the `viewport` maybe we should call the attribute `viewport`, instead of `size`? I think `size` is a bit confusing because we already [use it to define the actual size in the code](https://github.com/freenowtech/wave/blob/main/src/components/Headline/Headline.tsx):
```
    /**
     * Set the style of the headline
     */
    size?: 'xxl' | 'xl' | 'l' | 'm' | 's' | 'xs';
....
    switch (size) {
        case 'xxl':
            return h1Styles;
        case 'xl':
            return css`
                font-size: ${get('fontSizes.5')};
                line-height: 2.5rem;
            `;
        case 'l':
            return css`
                font-size: ${get('fontSizes.4')};
                line-height: 2rem;
            `;
```

so use something like `<Headline viewport={{_: 's', m: 'xl'}}>`?
And then we can use the same logic for the `Text` component,  in the future for spacers and other things that might depend on the viewport. Or I understand something wrong?

nlopin:
I'd like us to look at the component in isolation, without thinking about the UI the headline will be placed to. When using a component in the code, we describe the properties the component has. In that perspective current API provides clarity:
- `as` which tag to use
- `size` how big is the headline

If we replace `size` with `viewport` it doesn't provide any intuition, because it describes something outside of the component itself. 

This is the approach we follow through the whole design system (for margins, paddings, etc). A developer can set different values to a prop depending on the screen size, but it is not mandatory. This viewport thing is an optional trick they can use, but it is opt-out by default:

```js
<Headline size='l' mb={2}>Big title</Headline> // size for all the screens with margin-bottom fixed
<Headline size={{_: 'm', medium: 'xl'}} mb={{_: 0, medium: 2}}>Big title</Headline> // size and margin depends on the screen
```

If we use `viewport` instead of `size`, it loses the readability, especially in the default case (and we loose margin):

```js
<Headline viewport='l'>Big title</Headline> // size for all the screens
<Headline viewport={{_: 'm', medium: 'xl'}}>Big title</Headline> // depends on the screen
```

In my opinion, the size should stay, because we set size depending on the viewport, but not the other way around.

rafael-sepeda (approved):

src/components/Headline/docs/Headline.mdx
```diff
@@ -60,9 +59,24 @@ the semantics while following the design.
...
+
+```
+
+## Playground
+
 
 <Playground>
     <Headline as="h3">default headline h3</Headline>
     <Headline as="h1" size="xs">The smallest h1</Headline>
+    <Headline as="h4" size={{_: 'xs', medium: 'm'}}>Responsive headline</Headline>
```
div-Leo, line 81: Sorry for the question but even from the style-system is not 100% clear, I assume that `_` stands for the default value, this means that it will be `xs` unless the viewport size will be included in the medium breakpoints, is that right?

div-Leo (approved):

Issue #190, opened 2021-11-10
Box: BoxProps incompatible with Box component


vlad-nica:
-   `@freenow/wave` version: 1.16.1
-  `typescript` version: 4.4.4

### Relevant code

```js
import { Box, BoxProps } from '@freenow/wave'

const SomeComponent = (props: BoxProps) => {
    return <Box {...props} />
}
```

### What was expected to happen?

The code above throws a typescript error:
```
No overload matches this call.
  Overload 1 of 2, '(props: Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<...>> & { ...; } & BoxProps, "theme"> & Partial<...>, "theme"> & { ...; } & { ...; }): ReactElement<...>', gave the following error.
    Type '{ m?: ResponsiveValue<string | number | symbol, Required<Theme<TLengthStyledSystem>>> | undefined; margin?: ResponsiveValue<string | number | symbol, Required<...>> | undefined; ... 78 more ...; textAlign?: ResponsiveValue<...> | undefined; }' is not assignable to type 'Omit<Omit<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<HTMLDivElement>> & { ...; } & BoxProps, "theme"> & Partial<...>, "theme">'.
      Types of property 'color' are incompatible.
        Type 'ResponsiveValue<string | number | symbol, Required<Theme<TLengthStyledSystem>>> | undefined' is not assignable to type '(string & ResponsiveValue<string | number | symbol, Required<Theme<TLengthStyledSystem>>>) | undefined'.
          Type 'null' is not assignable to type '(string & ResponsiveValue<string | number | symbol, Required<Theme<TLengthStyledSystem>>>) | undefined'.
  Overload 2 of 2, '(props: StyledComponentPropsWithAs<"div", never, BoxProps, "theme", "div", "div">): ReactElement<StyledComponentPropsWithAs<"div", never, BoxProps, "theme", "div", "div">, string | JSXElementConstructor<...>>', gave the following error.
[...]

PR #197, opened 2022-01-10
Fix incompatible box props

jonotrujillo:
**What:**

- Avoid collision of `color` prop
- Add default theme declaration
- Use theme declaration in `Box` component
​
**Why:**

- Fix issue #190 
- Fix another issue where `Box` requires a value for `theme`

**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/src/components/Box/Box.spec.tsx
+++ b/src/components/Box/Box.spec.tsx
@@ -1,7 +1,7 @@
 import { render } from '@testing-library/react';
 import * as React from 'react';
 import { Colors } from '../../essentials';
-import { Box } from './Box';
+import { Box, BoxProps } from './Box';
 
 describe('Box', () => {
     it('renders without any props', () => {
@@ -31,6 +31,13 @@ describe('Box', () => {
     it('renders grid layout', () => {
         expect(
             render(<Box gridColumnGap={2} gridTemplateRows="1fr auto 10%" />).container.firstChild
+        ).toMatchSnapshot();
+    });
+
+    it('accepts props spreading', () => {
+        const renderBox = (props: BoxProps) => <Box {...props} />;
+        expect(
+            render(renderBox({ backgroundColor: Colors.POSITIVE_GREEN_900 })).container.firstChild
         ).toMatchSnapshot();
     });
 });
--- a/src/components/Box/Box.spec.tsx
+++ b/src/components/Box/Box.spec.tsx
@@ -35,9 +35,10 @@ describe('Box', () => {
     });
 
     it('accepts props spreading', () => {
-        const renderBox = (props: BoxProps) => <Box {...props} />;
-        expect(
-            render(renderBox({ backgroundColor: Colors.POSITIVE_GREEN_900 })).container.firstChild
-        ).toMatchSnapshot();
+        const renderBox = () => {
+            const props = { backgroundColor: Colors.POSITIVE_GREEN_900 };
+            return <Box {...props} />;
+        };
+        expect(renderBox).not.toThrow();
     });
 });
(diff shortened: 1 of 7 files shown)


src/components/Box/Box.tsx
```diff
@@ -24,13 +24,13 @@ interface BoxProps
     extends SpaceProps,
         LayoutProps,
         PositionProps,
-        ColorProps,
+        Omit<ColorProps, 'color'>, // HACK: avoid collision of `color` prop
```
nlopin, line 27: what else do we have inside colorProps in styled-system?

src/components/Box/Box.tsx
```diff
@@ -24,13 +24,13 @@ interface BoxProps
...
         FlexboxProps,
         GridProps,
         BackgroundProps,
         TextAlignProps {}
 
-const Box: StyledComponentBase<'div', never, BoxProps, 'theme'> = styled.div.attrs({ theme })<BoxProps>`
+const Box: StyledComponentBase<'div', DefaultTheme, BoxProps, 'theme'> = styled.div.attrs({ theme })<BoxProps>`
```
nlopin, line 33: Have you tried stripping off the whole `StyledComponentBase` type? I think Typescript can derive the correct type from the assignment

div-Leo (approved):

src/components/Box/Box.tsx
jonotrujillo: Then the project won't compile, because of other component that depend on `Box` breaking. I would like to have it this way, though. Looking into it now.

src/components/Box/Box.tsx
jonotrujillo: bg color and opacity

src/components/Box/Box.tsx
jonotrujillo: This is done now in c525f49 and 0405c22

src/components/Box/Box.spec.tsx
```diff
@@ -33,4 +33,11 @@ describe('Box', () => {
             render(<Box gridColumnGap={2} gridTemplateRows="1fr auto 10%" />).container.firstChild
         ).toMatchSnapshot();
     });
+
+    it('accepts props spreading', () => {
+        const renderBox = (props: BoxProps) => <Box {...props} />;
+        expect(
+            render(renderBox({ backgroundColor: Colors.POSITIVE_GREEN_900 })).container.firstChild
+        ).toMatchSnapshot();
```
nlopin, line 41: What is the idea behind making a snapshot? if the test checks prop spreading it should test only that the code is not crashing, no? 

Do you think we can remove the snapshot?

src/components/Search/Search.tsx
```diff
@@ -32,7 +34,11 @@ const ActiveBox = styled(Box)`
     }
 `;
 
-const SearchInputContainer = styled(Box)`
```
nlopin, line 35: I'm not quite getting why this change is here

nlopin (approved):
My main concern are snapshot test, let's discuss and merge

src/components/Box/Box.spec.tsx
jonotrujillo: No need for the snapshot, I can change the assertion to `not.toThrow()`. What do you think?

src/components/Search/Search.tsx
jonotrujillo: This changed is needed after removing the `StyledComponentBase` typing from `Box`. As `isInFocus` is not a prop in `BoxProps`, the interface is extended here.

PR #198, opened 2022-01-19
Add grid system components

jonotrujillo:
**What:**

Add grid system components.
​
**Why:**

In order to facilitate the creation of layouts that are adjusted to the 12 column grid used in our designs

The components can be used this way:

```
<Row mb={3}>
  <Column bg="#f1f2f4" textAlign="center" span={2}>
    1
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={2}>
    2
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={2}>
    3
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={2}>
    4
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={2}>
    5
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={2}>
    6
  </Column>
</Row>
<Row mb={3}>
  <Column bg="#f1f2f4" textAlign="center" span={3}>
    7
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={3}>
    8
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={3}>
    9
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={3}>
    10
  </Column>
</Row>
<Row mb={3}>
  <Column bg="#f1f2f4" textAlign="center" span={4}>
    11
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={4}>
    12
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={4}>
    13
  </Column>
</Row>
<Row mb={3}>
  <Column bg="#f1f2f4" textAlign="center" span={6}>
    14
  </Column>
  <Column bg="#f1f2f4" textAlign="center" span={6}>
    15
  </Column>
</Row>
<Row>
  <Column bg="#f1f2f4" textAlign="center" span={12}>
    16
  </Column>
</Row>
```

Will produce this layout:​

Screen Shot 2022-01-20 at 15 36 58

**How:**

Implementing `Row` and `Column` components

Currently supported:

- Column span
- Column offset

Not supported:

- Changing the gutter/gap between columns
- Responsive behaviour

**Checklist:**

-   [x] Add tests 
-   [x] Ready to be merged


rafael-sepeda (approved):
@jonotrujillo Looks cool and very clean! Thanks!

src/components/Grid/docs/Grid.mdx
```diff
@@ -0,0 +1,189 @@
...
+    </Column>
+    <Column bg="#f1f2f4" textAlign="center" span={2} offset={4}>
+      2
+    </Column>
+  </Row>
+</Playground>
+
+The `Row` and `Column` components can be used to create different kinds of layouts, combining them with other components of the design system.
+
+<Playground>
```
martimalek, line 155: Nice real world example!

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,29 @@
...
+import styled from 'styled-components';
+import { Spaces } from '../..';
+import { Box, BoxProps } from '../Box/Box';
+
+const COLUMN_WIDTH = 100 / 12;
+const GAP = Spaces[3];
+
+const Row = styled(Box).attrs({ display: 'flex', flexDirection: 'row', flexWrap: 'wrap' })`
+    gap: ${GAP};
+    margin-right: -${GAP};
```
nlopin, line 11: I think we can find a way to avoid setting negative margin-right. This is not a `gap` issue, because it sent only spacing between components in a flexbox container.

I suspect this is the effect of our usage of `calc` which assume that each column has a gap. This assumption is incorrect, the amount of gaps is the amount of columns minus one. If we fix the formula, we fix the issue

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,29 @@
...
+type ColumnOffset = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11;
+type ColumnSpan = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+
+interface ColumnProps extends BoxProps {
+    span?: ColumnSpan;
+    offset?: ColumnOffset;
+}
+
+const Column: FC<ColumnProps> = ({ span = 1, offset = 0, ...restProps }: ColumnProps) => {
+    const marginLeft = `${offset * COLUMN_WIDTH}%`;
```
nlopin, line 23: Have you considered using a [CSS grid](https://css-tricks.com/snippets/css/complete-guide-grid/) instead of a flexbox?

In my mind, we are trying to make CSS grid out of a flexbox container. All those things are supported in the `display:grid` nicely and we can be just a component that abstracts CSS props.

nlopin (changes requested):
Great work! I think a grid system can be useful for many developers, solving the challenge of wiring a layout CSS once and for good. 

I don't fully understand the decisions taken in this PR and would love to understand them better (see comments)

I also noticed that we don't touch the responsiveness of this component in the documentation. Do we have an intent to provide it?

src/components/Grid/Grid.tsx
jonotrujillo: I tried this in a previous iteration (244b393). What I did not like about it is that, to me, it did not made sense to obscure the grid api behind row and column components. I swathes the implementation to flex, as it scopes better that the contents are to be arranged in only one direction.

I really prefer the implementation suing grid, but then I would rework it in a way that we have `Grid` instead of `Row` and `GridItem` instead of `Column`

src/components/Grid/Grid.tsx
jonotrujillo: Makes sense. Let's check if we want to use grid or flex, and then decide if I should fix this.

martimalek: I don't have a strong opinion about it as long as we can use it to create a layout easily

src/components/Grid/Grid.tsx
martimalek: I don't have a strong opinion about it as long as we can use it to create a layout easily, even though it's true that it sounds weird to have a Grid system using flex instead of grid

martimalek (approved):
Very nice!

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const Column: FC<ColumnProps> = ({ span = 1, offset = 0, ...restProps }: ColumnProps) => {
+    const marginLeft = offset * COLUMN_WIDTH;
+    const width = span * COLUMN_WIDTH;
+
+    return (
+        <BorderBoxWrapper
+            flex={`0 0 ${width.toFixed(AMOUNT_OF_DECIMALS)}%`}
+            marginLeft={marginLeft ? `${marginLeft.toFixed(AMOUNT_OF_DECIMALS)}%` : undefined}
+            paddingRight={GAP}
```
nlopin, line 36: do you think padding is better than margin here?

nlopin (approved):

src/components/Grid/Grid.tsx
jonotrujillo: I have to use padding because I want the gap to be included in the width calculation. The alternative would be to use margin and set the width to `calc(width - gap)`.

PR #200, opened 2022-01-25
Feat/add missing icons

div-Leo:
**What:**
Update the icons with the missing ones and make them match with the design organization.
​Don't get afraid of the number of items, `src/icons/[type]` folders are generated by a script, and assets is containing just SVG changes.

**Why:**
We were missing some icons and I notice that most were in a different folder than they should be.

**How:**
- Added new icons and re-organized them into the correct folders.
- Generated the icons from the new assets
- Fix the docs with the new icons and divided them into sections
- Fix relative imports to export from the icons index file
​
**Media:**
Screenshot 2022-01-25 at 12-12-55 Icons

**Checklist:**

-   [ ] Needs review before the merge


nlopin:
I think now it is a breaking change.

We moved icons from their old places to new ones. That by default means the code will break for all users who used them in old locations (e.g. `import BasicIcons from '@freenow/wave/icons/basic' and then `<BasicIcons.DownloadIcon />` will break). Do you tackle that?

If we want to release it as a minor version, I suggest providing old exports too, but wrapping them into a `withDeprecatedMessage` function and pointing to the new place.

rafael-sepeda (approved):
@div-Leo All missing icons are now in the new list. Thanks!

div-Leo:
@nlopin isn't the default path to import icons `from '@freenow/wave'`? 
I was trying to use the import accessing directly the category of icons and it gives me an error, also in the docs it doesn't mention anything about the "categories".

If it was not, I think we should enforce it to avoid any future breaking change just by moving the icons' location.

martimalek (approved):
Amazing job! As long as this does not break the icons imports then for me it's great 😁

nlopin (approved):
All good @div-Leo 👍

I've checked our codebase and couldn't find any direct imports of icons. Just double-check that you do not delete existing icons (if so, they need to be copied to the deprecated folder) and merge

div-Leo:
Thanks @nlopin! 

I checked and made sure that I didn't remove any icon. 👍 

I'm going to proceed with the merge!

PR #201, opened 2022-01-25
Allow to select page size in Pagination

### LINKED ISSUE #199: Pagination can display the page size and select a different one

I'd like the `Pagination` component to optionally display the **page size** and allow to change this value selecting it from a given set of values.

In order to select the wanted page size a `SelectList` would be used:
Screenshot 2022-01-21 at 13 43 42

**Reasoning**

The reason why I created this issue is because in GAME we want to allow our users to select how many quests they want to see in the quest list from a given set. The `Pagination` is being used there but it currently lacks this feature.

**New API**
I'd add the following props
```jsx
interface PaginationProps {
    /**
     * Page sizes to choose from.
     */
    pageSizes?: {label: string, value: string}[];
    /**
     * Callback when a different page size was selected. Can only be used when `pageSizes` is passed.
     */
    onSelectPageSize?: (selected: { label: string, value: string }) => void;
    ...
}
```

martimalek:
**What:**

Allows to change the page size when using the `Pagination` component.

​
**Why:**

Closes #199 

​
**How:**

- Add a `SelectList` on the left of the `Pagination` buttons which displays a set of values received by props
- Add `onSelectPageSize` prop to handle selecting the page size
​
**Media:**

Screenshot 2022-01-25 at 13 54 19

**Checklist:**

-   [x] Make select container aria-label a prop
-   [x] Ready to be merged

**Comments:**

In order to test the select behaviour I've had to install `react-select-event`.


package.json
```diff
@@ -95,6 +95,7 @@
         "react-dom": "^16.13.1",
         "react-helmet": "^6.0.0",
         "react-lottie": "^1.2.3",
+        "react-select-event": "^5.3.0",
```
martimalek, line 98: Installed in order to test the select behaviour

src/components/Pagination/Pagination.tsx
```diff
@@ -101,15 +118,26 @@ const Pagination: React.FC<PaginationProps> = ({
...
     const isFirstPage = value === 1;
     const isLastPage = totalItems > 0 ? value === Math.ceil(totalItems / pageSize) : true;
+    const hasMultiplePageSizes = pageSizes.length > 0;
 
     return (
         <Container>
-            <div>
+            <ButtonsContainer>
+                {hasMultiplePageSizes && (
+                    <Box aria-label="Select page size container" position="absolute" left="0" width="4.5em">
```
martimalek, line 133: I will make this _aria-label_ come from a prop so that it is more flexible

src/components/Pagination/docs/PaginationProvider.tsx
```diff
@@ -7,54 +7,77 @@ interface Props {
...
         handleSkipBackward: () => void;
     }) => ReactNode;
 }
 
 interface State {
     currentPage: number;
+    pageSize: number;
 }
 
 class PaginationProvider extends React.Component<Props, State> {
```
div-Leo, line 25: I know it's an example, but can we take advantage of the changes to make it a function?  😊

div-Leo (commented):
Hey @martimalek, very nice job! 
I have to be honest and say that still, I don't see the advantage of having it bound to the pagination. 
It's just adding a SelectList based on a boolean but still controlled by the parent, where most of the implementation will be anyway. On the contrary, I see the limitation on the position of the SelectList that can be only placed to the left, unless we pass another prop.

I can see that it is related to the Pagination, so if the others agree on making these changes I'm totally fine!

src/components/Pagination/docs/PaginationProvider.tsx
martimalek: Yes sure, I'll have it done by this afternoon

martimalek: Done 👍🏽

nlopin:
On wave sync we decided to discuss it again when @rafael-sepeda is back to understand the reusability aspect from the design side better

div-Leo:
When we fix the conflicts, as agreed, we can proceed with this PR

nlopin (approved):

rafael-sepeda (approved):

PR #208, opened 2022-02-24
Remove brand colors from semantic tokens

nlopin:
## What

As we discussed on the sync, we want to separate branding colors from semantic tokens
​
**How:**

* remove `brand` section of the semantic color object (I searched the codebase, it is not used)

**Checklist:**

-   [x] Ready to be merged

Since it is not used in the codebase, let's make it a minor release even though technically it is a breaking change.


rafael-sepeda (approved):
We will need, in the future, to adjust the Naming convention for the brand colors. For now it is clear that we can't do that.

PR #214, opened 2022-04-11
Different children with the same key in Datepicker

martimalek:
**What:**

Use generated ids as keys in Datepicker

​
**Why:**

When using the Datepicker component a warning appeared when the Datepicker opened due to multiple keys having the same value in the component that represents each day of the month. Check the screenshots to understand how the keys' values were.

​
**How:**

- Generate an id for each day to ensure that the values are different

​
**Media:**

image
Screenshot 2022-04-11 at 17 09 41

**Checklist:**

-   [ ] Ready to be merged

### DIFF
--- a/src/components/Datepicker/Month.tsx
+++ b/src/components/Datepicker/Month.tsx
@@ -3,6 +3,7 @@ import { useMonth, FirstDayOfWeek } from '@datepicker-react/hooks';
 import styled from 'styled-components';
 import { format } from 'date-fns';
 
+import { generateId } from '../../utils/ids';
 import { Colors } from '../../essentials';
 import { Text } from '../Text/Text';
 import { Day } from './Day';
@@ -61,7 +62,7 @@ const Month: FC<MonthProps> = ({ year, month, firstDayOfWeek, locale }: MonthPro
                         return <Day date={day.date} key={day.date.toString()} dayLabel={day.dayLabel} />;
                     }
 
-                    return <div key={day} />;
+                    return <div key={generateId()} />;
                 })}
             </DaysContainer>
         </div>


src/components/Datepicker/Month.tsx
```diff
@@ -61,7 +62,7 @@ const Month: FC<MonthProps> = ({ year, month, firstDayOfWeek, locale }: MonthPro
                         return <Day date={day.date} key={day.date.toString()} dayLabel={day.dayLabel} />;
                     }
 
-                    return <div key={day} />;
+                    return <div key={generateId()} />;
```
div-Leo, line 65: This fixes the warning but doesn't solve the issue 100% from the React perspective, I guess it's not crazy computation, but it will rerender all the cells every time because is a new id every time.

What about using indexes? is going to receive the same index for the first empty cells, until it gets the date for the actual first day of the month. It's a bit ugly I guess, but would work in this case I think.

Maybe I'm being too picky and it wouldn't add any value, what do you think?

martimalek: I'm not sure the key would affect the amount of rerenders since there are no changes to the order of the items in the list, the only moment where it will have to rerender is when the parent rerenders (when the Datepicker is opened or the user changes the month). 
But I do think as well that for this case indexes would work, since the order does not change the indexes can always uniquely identify the items. I used a generated id because this is the foolproof way of setting a key to a list item when the data does not already have an id, but if everyone agrees I can change it to use the indexes.

src/components/Datepicker/Month.tsx
```diff
@@ -61,7 +62,7 @@ const Month: FC<MonthProps> = ({ year, month, firstDayOfWeek, locale }: MonthPro
                         return <Day date={day.date} key={day.date.toString()} dayLabel={day.dayLabel} />;
                     }
 
-                    return <div key={day} />;
```
nlopin, line 64: do we understand why the day doesn't work? shouldn't we just create a timestamp for those dates and use them as a key? They are stable

martimalek: The issue is that when a month does not start on monday the calendar renders empty cells to represent the days of the last month included in the current month's calendar (check out the second image in the description). In that case `day` is not a date and we cannot use it as a unique key (I don't remember if `day` equals 0 or `undefined`, but either way we cannot use it).

div-Leo (approved):

nlopin:
@martimalek do we merge?

Issue #215, opened 2022-04-12
Export custom hooks to extract the logic outside


div-Leo:
### Export custom hooks to extract the logic outside

Provide specific types of hooks from the library so to extract the logic outside.

Here for example the Toggle component could use the `useToggle` function to ad the on/off logic but it can be used also in other scenarios, so we need to write it every time.

```js
import { useReducer } from 'react';
​
export type DispatchWithOptionalAction<Type> = (_arg?: Type | unknown) => void;
​
export type UseToggleType<Type> = [Type, DispatchWithOptionalAction<Type>];
​
function _toggle(currentValue: boolean, newValue: boolean | undefined) {
  return typeof newValue === 'boolean' ? newValue : !currentValue;
}
​
function useToggle(initialValue: boolean = false): UseToggleType<boolean> {
  return useReducer(_toggle, initialValue);
}
​
export default useToggle;
```
And this would be the usage.

```js
import { useToggle, Toggle } from '@freenow/wave'

function Component () {
    const [checked, toggle] = useToogle(false)
    return (
        <>
            <Toggle checked={checked} onChange={toggle} />
        </>
    )
}
```

PR #220, opened 2022-05-03
Feat/export use toggle hook

div-Leo:
Linked issue: #215 

## Add `useToggle` custom hook and export hooks from Wave

Added a new `useToggle` hook to provide out-of-the-box the logic that toggles a boolean. It can be used in toggle implementation as well as internally in the library, e.g. accordion component logic to toggle open-close state.

This PR also aims to export the custom hooks outside the library so to be used if necessary.


rafael-sepeda (approved):

martimalek (commented):
I would enforce having unit tests for every hook if we are going to expose them, wdyt?

This PR is probably a good place to add the tests for the `useToggle`, as for the others, if we agree we can create issue/s for them.

src/utils/hooks/useToggle.tsx
```diff
@@ -0,0 +1,15 @@
+import { useReducer } from 'react';
+
+export type DispatchWithOptionalAction<Type> = (_arg?: Type | unknown) => void;
+
+export type UseToggleType<Type> = [Type, DispatchWithOptionalAction<Type>];
+
+function toggle(currentValue: boolean, newValue: boolean | undefined) {
```
nlopin, line 7: It doesn't look right to me. Can you explain the use case here? If we can set the value manually, it shouldn't be called `useToggle` then


The API I expected from the `useToggle` is:
- it takes a boolean as the initial value `useToggle(true|false)`
- it exposes the current value
- it exposes a function to toggle from the current state to the opposite: `toggle()`

nlopin (changes requested):
My summary is:
* let's discuss the API because I don't understand why we call it `useToggle` at the moment
* `useToggle` hook differs from `utils` hook which contains purely internal details. `useToggle` sounds like a whole component logic we expose for others, so it should be a separate folder on the `src` level (maybe call it `hooks`)
* I love the idea @martimalek mentioned on testing the behaviour, unless it is not clear (which I think in this case can be)

martimalek:
> * `useToggle` hook differs from `utils` hook which contains purely internal details. `useToggle` sounds like a whole component logic we expose for others, so it should be a separate folder on the `src` level (maybe call it `hooks`)

I agree with @nlopin, makes sense to have a separate `hooks` folder for this. 

Also, if the other hooks are only supposed to be used internally, maybe we don't even need to expose them... What about only exposing `useToggle` in this PR and then discuss if the others are a good enough use case?

div-Leo:
To avoid adding too many comments I'll answer all the questions here, trying to be brief:
- I'll add tests for `useToggle` 👍
- I can add them also for any other hook we will decide to expose
- I did think it was better to have a `./hook` folder but didn't want to make too much noise.
- I don't agree when you say that they are "purely internal details", why can't we expose such functionalities that can be used outside, instead of having to reimplement them in every project. e.g. the `useControlledState` can be nice to expose. 
- [here](https://github.com/freenowtech/wave/issues/215) shows the use case, the idea is to abstract a logic that can add a "reactive" state that provides a value (boolean) and a function that toggle that value.

src/utils/hooks/useToggle.tsx
```diff
@@ -0,0 +1,15 @@
...
+    return typeof newValue === 'boolean' ? newValue : !currentValue;
+}
+
+function useToggle(initialValue = false): UseToggleType<boolean> {
+    return useReducer(toggle, initialValue);
```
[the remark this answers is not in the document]
div-Leo: I find this implementation cleaner than having to wrap the setter function in another abstraction. Moreover, we can declare the function outside instead of having to redeclare every time. 

Why do you think it adds extra complexity? 
I do agree that it might be more complicated to read as we are probably less used to using `useReducer` than a simple `useState`.

src/utils/hooks/useToggle.tsx
div-Leo: I guess the idea is to let the user force a state.

To also extend the idea of usage, imagine we have a bunch of expendable components and we want to close them all we can call `toggle(false)` instead of checking each value and if it's true then we close it.
Maybe not the best example but I hope it can be clearer.

I guess the API would be the same with a small correction: 
- it takes a boolean as the initial value `useToggle(true|false)`
- it exposes the current value
- it exposes a function to toggle from the current state to the opposite: `toggle()` || passes the boolean to set `toggle(true)`

nlopin: What do you think about exposing another function to enforce the state? Let's say `setToggle` to make it clear

nlopin:
> useControlledState can be nice to expose.

`useControlledState` is used to showcase examples in the documentation. I think before exposing hooks we need to answer the question of why a design system should expose this hook? Why do we expose hooks at all?

div-Leo:
> `useControlledState` is used to showcase examples in the documentation
It's not. I added that hook myself when I wanted to make the `Search` component a controlled component.

>  Why do we expose hooks at all?
IMO, it makes sense that a Components library provides presentationals and UI-related logic. Certain types of logic might repeat over and over again in the different projects and having it coming from the library might speed up the development, or at least standardize the behaviors.

nlopin:
Don't get me wrong, there is a lot of useful code, but if we want to export all the useful code we have, it won't be a design system library anymore. It will be a library with useful stuff.

I see a primary goal of the wave package is to give ready-to-use components and as a secondary thing hooks to reimplement custom components but be confident about the behavior. The `useControlledState` doesn't provide anything from that list. 

I see the value in having often used hooks, maybe we should consider creating a separate library for that

src/utils/hooks/useToggle.tsx
div-Leo: meaning returning something like this?
```js
return [value, toggle, setToggle]
```

div-Leo:
It could be nice to have it structured in sub reops like `wave/hook` or `wave/components`. But I guess it would be a breaking change and too much work at the moment.

src/utils/hooks/useToggle.tsx
nlopin: Yep, or even just set since we are in the context of the toggle

src/utils/hooks/useToggle.tsx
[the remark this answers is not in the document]
nlopin: but if you return the reducer, it means, we return [dispatch, state]. What usage will be here?

nlopin:
@div-Leo do you want me to finish the PR?

nlopin (approved):

martimalek (approved):

PR #221, opened 2022-05-03
feature: Add Divider component to Wave

### LINKED ISSUE #218: New component: Divider
A number of our tools that use Wave DS, including BOMT, DMT, PMT (and possibly more) have many instances of a horizontal or vertical line separating pieces of content. It is being redefined over and over, sometimes as border, sometimes as a div element, despite the fact that its definition is (almost) always identical: 
- **1 pixel width**
- **solid** border
- **color**: AUTHENTIC_BLUE_200 = '#C6CDD4', or in semantic colors **$border.primary**

---

_Examples_

**Navigation Bar (DMT)**
Screenshot 2022-05-02 at 16 23 49

**Advanced Search (DMT)**
Screenshot 2022-05-02 at 16 24 35

**Fare Breakdown (BOMT)**
Screenshot 2022-05-02 at 16 26 35

---

A **Divider** component is often a core component in design systems, therefore, this issue proposes the addition of new <Divider> component to our Wave design system for these and other use-cases.

---

### Proposed architecture

Screenshot 2022-05-02 at 16 31 02

_Note: Little correction, the color would be retrieved as semanticColor $border.primary not from Colors_

JanHamara:
**What:**

This PR adds a new Divider component to the Wave Design System, as per: [#218](https://github.com/freenowtech/wave/issues/218)

​
**Why:**
Closes #218 

A number of our tools that use Wave DS, including BOMT, DMT, PMT (and possibly more) have many instances of a horizontal or vertical line separating pieces of content. A Divider component is often a core component in design systems, therefore, this issue proposed the addition of new component to our Wave design system for these and other use-cases.
​
​
**Media:**
Screenshot 2022-05-03 at 17 33 35
Screenshot 2022-05-03 at 17 33 51

**Checklist:**
-   [x] Reviewed with designer
-   [x] Ready to merge

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -9,13 +9,12 @@
   ],
   "rules": {
     "declaration-empty-line-before": null,
-    "unit-whitelist": [
-      "rem",
-      "deg",
-      "fr",
-      "ms",
-      "%"
-    ],
+    "declaration-property-unit-whitelist": {
+      "/.*/": ["rem", "deg", "fr", "ms", "%", "px"]
+    },
+    "declaration-property-value-blacklist": {
+      "/.*/": ["(\\d+[1]+px|[^1]+px)"]
+    },
     "value-keyword-case": ["lower", { "ignoreKeywords": ["dummyValue"] }]
   }
 }
--- a/src/components/Divider/Divider.tsx
+++ b/src/components/Divider/Divider.tsx
@@ -18,59 +18,31 @@ interface DividerProps extends SpaceProps {
 }
 
 const HorizontalLine: StyledComponent<'div', typeof theme, DividerProps, 'theme'> = styled.div.attrs({ theme })<
-    DividerProps
+    Pick<SpaceProps, 'my'>
 >`
     width: 100%;
-    margin: 0 auto;
+    margin-left: auto;
+    margin-right: auto;
     border: 0;
-    border-top: 0.06rem solid ${get('semanticColors.border.primary')};
+    border-top: 1px solid ${get('semanticColors.border.primary')};
 
     ${compose(space)}
 `;
 
 const VerticalLine: StyledComponent<'div', typeof theme, DividerProps, 'theme'> = styled.div.attrs({ theme })<
-    DividerProps
+    Pick<SpaceProps, 'mx'>
 >`
     display: inline-block;
     width: 0.06rem;
-    margin: 0 auto;
+    margin-top: 0;
+    margin-bottom: 0;
     border: 0;
-    border-left: 0.06rem solid ${get('semanticColors.border.primary')};
+    border-left: 1px solid ${get('semanticColors.border.primary')};
 
     ${compose(space)}
 `;
 
-/**
- * ** Primary UI element for visually separating content **
- *
- * Renders a divider UI component: horizontal or vertical line that visually separates two pieces of data, content or UI immediately next to it.
- *
- * Horizontal divider will take up full available width, vertical divider will take up full available height.
- *
- * _Divider_ renders a horizontal divider by default. Set **vertical** prop to __true__ to change divider orientation.
- *
- *
- * ---
- *
- * <br/>
- *
- * #### Divider vs. Border
- *
- * The default color of _Divider_ is $border.primary (#C6CDD4) <span style="color: #C6CDD4">⬤</span>, however...
- *
- * ** Divider is NOT a border, and should not be used as such. Please do not use this component as a border for elements. **
- *
- * Divider is naturally expected to have a certain offset from the elements it is 'dividing' or separating.
- *
- * ---
- *
- * #### Style Props
- *
- * The Divider has following design props:
- * - **offset** - set the divider offset from the content it is separating (uses _mx_, _my_ styled system props)
- *
- */
-const Divider = ({ vertical = false, offset = 0 }: DividerProps): React.ReactElement =>
+const Divider: React.FC<DividerProps> = ({ vertical = false, offset = '1rem' }: DividerProps) =>
     vertical ? (
         <VerticalLine mx={offset} data-testid="vertical-divider" />
     ) : (
--- a/src/components/Divider/Divider.tsx
+++ b/src/components/Divider/Divider.tsx
@@ -1,0 +1,80 @@
+import React from 'react';
+import styled, { StyledComponent } from 'styled-components';
+import { compose, space, SpaceProps } from 'styled-system';
+import { theme } from '../../essentials/theme';
+import { get } from '../../utils/themeGet';
+
+type DividerOffset = number | string;
+
+interface DividerProps extends SpaceProps {
+    /**
+     * Set the direction of the divider to vertical
+     */
+    vertical?: boolean;
+    /**
+     * Set offset / margin of the divider from the surrounding content
+     */
+    offset?: DividerOffset;
+}
+
+const HorizontalLine: StyledComponent<'div', typeof theme, DividerProps, 'theme'> = styled.div.attrs({ theme })<
+    DividerProps
+>`
+    width: 100%;
+    margin: 0 auto;
+    border: 0;
+    border-top: 0.06rem solid ${get('semanticColors.border.primary')};
+
+    ${compose(space)}
+`;
+
+const VerticalLine: StyledComponent<'div', typeof theme, DividerProps, 'theme'> = styled.div.attrs({ theme })<
+    DividerProps
+>`
+    display: inline-block;
+    width: 0.06rem;
+    margin: 0 auto;
+    border: 0;
+    border-left: 0.06rem solid ${get('semanticColors.border.primary')};
+
+    ${compose(space)}
+`;
+
+/**
+ * ** Primary UI element for visually separating content **
+ *
+ * Renders a divider UI component: horizontal or vertical line that visually separates two pieces of data, content or UI immediately next to it.
+ *
+ * Horizontal divider will take up full available width, vertical divider will take up full available height.
+ *
+ * _Divider_ renders a horizontal divider by default. Set **vertical** prop to __true__ to change divider orientation.
+ *
+ *
+ * ---
+ *
+ * <br/>
+ *
+ * #### Divider vs. Border
+ *
+ * The default color of _Divider_ is $border.primary (#C6CDD4) <span style="color: #C6CDD4">⬤</span>, however...
+ *
+ * ** Divider is NOT a border, and should not be used as such. Please do not use this component as a border for elements. **
+ *
+ * Divider is naturally expected to have a certain offset from the elements it is 'dividing' or separating.
+ *
+ * ---
+ *
+ * #### Style Props
+ *
+ * The Divider has following design props:
+ * - **offset** - set the divider offset from the content it is separating (uses _mx_, _my_ styled system props)
+ *
+ */
+const Divider = ({ vertical = false, offset = 0 }: DividerProps): React.ReactElement =>
+    vertical ? (
+        <VerticalLine mx={offset} data-testid="vertical-divider" />
+    ) : (
+        <HorizontalLine my={offset} data-testid="horizontal-divider" />
+    );
+
+export { Divider, DividerProps, DividerOffset };
--- a/src/components/Divider/docs/Divider.mdx
+++ b/src/components/Divider/docs/Divider.mdx
@@ -31,7 +31,7 @@ Divider is naturally expected to have a certain offset from the elements it is '
 <br/>
 
 ### Style Props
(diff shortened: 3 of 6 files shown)


src/components/Divider/docs/Divider.mdx
```diff
@@ -0,0 +1,88 @@
...
+- **offset** - set the divider offset from the content it is separating (uses _mx_, _my_ styled system props)
+
+<hr/>
+
+## Usage 
+
+### Horizontal (without offset)
+
+<ItemWrapper>
+    <WrappedHorizontalDivider offset={0} />
```
div-Leo, line 49: Does `offset={0}` affect the result? meaning if I don't pass the offSet how would it look like?

src/components/Divider/docs/Divider.mdx
```diff
@@ -0,0 +1,88 @@
...
+<hr/>
+
+### Style Props
+
+The Divider has following design props:
+- **offset** - set the divider offset from the content it is separating (uses _mx_, _my_ styled system props)
+
+<hr/>
+
+## Usage
```
div-Leo, line 44: Can we add an example using the `<Playground>`? I see it's imported but I can't see it in the docs.

src/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * ---
+ *
+ * #### Style Props
+ *
+ * The Divider has following design props:
+ * - **offset** - set the divider offset from the content it is separating (uses _mx_, _my_ styled system props)
+ *
+ */
+const Divider = ({ vertical = false, offset = 0 }: DividerProps): React.ReactElement =>
```
div-Leo, line 73: @rafael-sepeda would be better to have a default margin as most of the time might be the same or it's going to change and so it makes sense to manually pass it every time?

src/components/Divider/docs/Divider.mdx
```diff
@@ -0,0 +1,88 @@
...
+```jsx
+<Divider />
+```
+
+<hr/> 
+
+### Horizontal (with offset)
+
+<ItemWrapper>
+    <WrappedHorizontalDivider offset={5} />
```
div-Leo, line 61: I think I'd like to see an example with a margin-top different from the bottom as it's the way I saw it used the most.

src/components/Divider/Divider.tsx
rafael-sepeda: @div-Leo  From my perspective, we should always have default values from props.

div-Leo: @rafael-sepeda what would be the default spaceing?

src/components/Divider/docs/Divider.mdx
JanHamara: Sure!

src/components/Divider/docs/Divider.mdx
JanHamara: Yeah, good point, I can make it in a way so we don't need to pass 0.

nlopin:
@JanHamara if it is not ready to review, you can make a PR draft

src/components/Divider/Divider.tsx
JanHamara: @div-Leo default offset (spacing) is 0 right now, so none :)

src/components/Divider/docs/Divider.mdx
JanHamara: We're still having active discussion about how margins will work on the issue, however, I'm not sure if we will even manage top and bottom margin separately for Divider: https://github.com/freenowtech/wave/issues/218.

Could you link an example or screenshot of that case you mentioned, meanwhile please? Is it really a Divider if it has different offset from the two items it is positioned between? Is it not more of a border, rather?

rafael-sepeda: @div-Leo , @JanHamara in design we use a lot of different top/bottom margins. But since we have in addition spacers we can simplify the dividers and consider having the same top/bottom spacing and in the case when we need to have a different from one of the sides - use an additional spacer.

src/components/Divider/Divider.tsx
div-Leo: Yep, I know, my suggestion is to have some significant default value

src/components/Divider/docs/Divider.mdx
div-Leo: I saw spacers used in the layout but I don't think we have the equivalent programmatically (in Wave).

div-Leo: > Could you link an example or screenshot of that case you mentioned, meanwhile please? Is it really a Divider if it has different offset from the two items it is positioned between? Is it not more of a border, rather?

Check "Advanced search" form, we have Dividers with different margin top/bottom
Screenshot 2022-05-04 at 14 30 23

src/components/Divider/Divider.tsx
JanHamara: No problem for me to set some default offset, idk where we go from to decide what it should be tho ❓ 

I guess we could take it from these two use cases?
Screenshot 2022-05-04 at 15 32 08

The offset of Dividers is also same as spacing between LabelValue items in the block (in the column flow) = **1rem**

I believe we can safely assume the divider offset will always be same in this kind of block and set default offset to **1rem**.

src/components/Divider/docs/Divider.mdx
JanHamara: I agree with Rafael's solution, I think the example in Advanced Search you linked it isn't really a goal to have a divider with different margins, if you look at the other collapsible items it's visible that there is some additional whitespace after the form (like above Hailing Type section above the highlighted yellow area) and because of that whitespace there is a smaller top margin, because we actually want to have same spacing both top and bottom of the line. So in this case, the additional whitespace above yellow area + top-margin **is (almost) same as** bottom-margin. 

So normally, we wouldn't have any padding after the select input field, and we would have a divider that equally separates both sections (as according to couple design and composition principles like alignment and proximity).

@rafael-sepeda correct me if I'm wrong, but that's the composition I would expect, unless there is particular reason why one section would have different offset from divider than the other?

Screenshot 2022-05-04 at 15 52 30

div-Leo: "(almost)" still means different...  Honestly, I don't get what you're trying to say. What you say is the top margin is the bottom margin of the input, I guess what matters is what we have in the code not what we see... 

Anyway, this is something that only @rafael-sepeda can answer, so I'm going to leave it to you guys.

rafael-sepeda: > I saw spacers used in the layout but I don't think we have the equivalent programmatically (in Wave).

@div-Leo we do have them in Wave:
https://github.com/freenowtech/wave/tree/main/src/essentials/Spaces
https://wave.free-now.com/essentials/spaces

In Figma we also have components for them that use same specs.

rafael-sepeda: > correct me if I'm wrong, but that's the composition I would expect, unless there is particular reason why one section would have different offset from divider than the other?
Sometimes we need to have different spacings from dividers if the dividers show the end of one section and you need to have a bigger spacing after to show a more clear start for the next section. But more often the top/bottom spacing is the same, especially if this is part of a component or a list of components, so, from my perspective for simplicity, we can set one margin value for top/bottom for spacers in dividers and then use additional spacers if needed.

src/components/Divider/Divider.tsx
JanHamara: @div-Leo I will set the default spacing to **1rem** then, or do you have other suggestion?

src/components/Divider/docs/Divider.mdx
JanHamara: Ok, thank you @rafael-sepeda, I will **resolve** this conversation.

src/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,80 @@
...
+`;
+
+const VerticalLine: StyledComponent<'div', typeof theme, DividerProps, 'theme'> = styled.div.attrs({ theme })<
+    DividerProps
+>`
+    display: inline-block;
+    width: 0.06rem;
+    margin: 0 auto;
+    border: 0;
+    border-left: 0.06rem solid ${get('semanticColors.border.primary')};
```
nlopin, line 38: We agreed to change it to `1px` because the divider shouldn't scale with font size

div-Leo (approved):

rafael-sepeda (approved):

src/components/Divider/docs/Divider.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+**Primary UI element for visually separating content**
+
+Renders a divider UI component: horizontal or vertical line that visually separates two pieces of data, content or UI.
+<br/>
+
+### Default Behaviour
+Horizontal divider takes up full available width, vertical divider takes up full available height.
+
+Horizontal divider is rendered by default. Set **vertical** prop to __true__ to change divider orientation.
```
martimalek, line 22:
```suggestion
The horizontal divider is rendered by default. Set **vertical** prop to `true` to change the divider orientation.
```

src/components/Divider/docs/Divider.mdx
```diff
@@ -0,0 +1,101 @@
...
+### Divider vs Border
+The default color of Divider is **$border.primary** (#C6CDD4), however...
+
+**Divider is NOT a border, and should not be used as such. Please do not use this component as a border for elements.**
+
+Divider is naturally expected to have a certain offset from the elements it is 'dividing' or separating.
+<br/>
+
+### Style Props
+The Divider has following design props:
```
martimalek, line 34:
```suggestion
The Divider has the following design props:
```

martimalek (approved):
I've left some small comments, mostly wording. Overall great job!

.stylelintrc
```diff
@@ -9,13 +9,12 @@
...
-      "rem",
-      "deg",
-      "fr",
-      "ms",
-      "%"
-    ],
+    "declaration-property-unit-whitelist": {
+      "/.*/": ["rem", "deg", "fr", "ms", "%", "px"]
+    },
+    "declaration-property-value-blacklist": {
```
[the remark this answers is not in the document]
JanHamara: to only allow '1px' not any pixel value :)

src/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,52 @@
...
+    width: 0.06rem;
+    margin-top: 0;
+    margin-bottom: 0;
+    border: 0;
+    border-left: 1px solid ${get('semanticColors.border.primary')};
+
+    ${compose(space)}
+`;
+
+const Divider: React.FC<DividerProps> = ({ vertical = false, offset = '1rem' }: DividerProps) =>
```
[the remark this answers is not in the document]
JanHamara: I kept it because we type most Wave components as React.FC, otherwise I wouldn't really put any type here :) There is loads of articles explaining why React.FC should not be really used anymore, e.g. https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/

.stylelintrc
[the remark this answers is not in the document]
martimalek: Awesome!

src/components/Divider/Divider.tsx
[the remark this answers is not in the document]
martimalek: I wouldn't block the merge of this PR because of this, but maybe it's a good moment to stop adding return types to new components.
Check out [this Kent C. Dodds article](https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript), particularly this bit:

> Ok, so what about the return value? Well, we could type it as React.ReactElement or even wider as a JSX.Element. But honestly, I side with my friend [Nick McCurdy](https://twitter.com/nickemccurdy) when [he says](https://twitter.com/nickemccurdy/status/1365384372908621833) that mistakes can easily be made causing the return type to be too wide. So even outside a react context, I default to not specifying the return type (rely on inference) unless necessary. And that's the case here.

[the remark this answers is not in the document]
martimalek: Feel free to resolve this if you want to merge and I'll open a new issue to have the discussion there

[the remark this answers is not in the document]
JanHamara: great, I totally agree with you here, I will remove the return type for this PR too!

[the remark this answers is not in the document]
JanHamara: Totally agree with you, I would remove it on this PR, but it is currently enforced by eslint, so let's open a separate issue first as you say, to make the proposal :)

PR #222, opened 2022-05-06
Update `react-windowed-select` in order to solve "window is undefined" in Next.js

### LINKED ISSUE #209: Wave is crashing in next.js projects
-   `@freenow/wave` version: 1.15.1 and later

Any project that uses the [`SelectList` component ](https://wave.free-now.com/components/select-list).

image

It is caused by https://github.com/jacobworrel/react-windowed-select/issues/74

### Reproduction

TBD

vitalybaev:
**What:**

This package cannot be used in Next.js projects because of error "window is undefined".

**Why:**

Closes #209
​
**How:**

Wave uses `react-windowed-select`, which uses `react-select`. "window is undefined" actually occurs in `react-select`, but was resolved and backported to [`react-windowed-select`](https://github.com/jacobworrel/react-windowed-select/issues/62).

Currently, Wave uses `react-windowed-select@2x` which points to `react-select@3.x`. Meanwhile, `react-windowed-select@3.x` which fixed this problem  uses `react-select@4.x`.

Despite major version has been increased, I believe that upgrade is seamless. Take a look at the [release notes of react-select@4.0.0](https://github.com/JedWatson/react-select/releases/tag/react-select%404.0.0). The only breaking change is emotion related: "The NonceProvider component now requires a cacheKey prop that corresponds to the key for the Emotion cache.".

Another things that `react-windowed-select@3x` [shipped are](https://github.com/jacobworrel/react-windowed-select/blob/master/CHANGELOG.md#300---2021-01-23):

* react-window patch version update
* peer deps to add support for React 17
* dev dep minor/patch versions

Also, I don't see some visual regression with `SelectList` component, and tests have passed.

**Checklist:**
-   [x] Ready to be merged


vitalybaev:
Hmmm, it breaks documentation building ([as here](https://github.com/jacobworrel/react-windowed-select/issues/74))

vitalybaev:
Converted to Draft since there are bundling issues in `react-windowed-select` that must be fixed before

nlopin:
Do you see any other option? The bundling issue is there for quite a long time and I was considering either to fix it or just migrate to something more supported

vitalybaev:
@nlopin I've created a [pull request](https://github.com/jacobworrel/react-windowed-select/pull/91) to the library that fixes this issue.
But of course we don't know when it will be merged.
Another reason is to publish your own package within `@freenow/react-windowed-select` with the fix until it will be merged in the upstream repository.

vitalybaev:
I have good news and bad news.

Good news is that my PR in `react-windowed-select` has been merged.
Bad news is that right after that PR with upgrading `react-select` to v5 has been also merged and both these changes have been released at once.
The problem is that v5 version of `react-select` came with a lot of changes:
* Dropped IE11 support. As far as I know, Wave supports IE11.
* Completely rewritten with TypeScript + some breaking changes. I've faced a lot of TS errors when tried to build Wave with the package. Of course that's not the problem, but I decided not to spent my time solving it without clarifying IE11 status.

The good news that author of `react-windowed-select` released `5.0.0` after `3.x`, so it's possible to release only my fix as `4.0`. I asked the author about that. Let's wait for his reponse.

vitalybaev:
I've got some news here!

The author of the `react-windowed-select` published intermediate version of the package (with my bundling fix, but before upgrading to `react-select@5`).

The package now works fine in Next.js (and other Node.js environments such as Gatsby or so) 🎉

But there are some additional problems, that I tried to fix:

1. `disabledStyles.control(...)` at `src/components/SelectList/styles.ts:7` receives the only argument - `SelectListProps`, but it's called this way: `disabledStyles.control({ isFocused: state.isFocused, ...state.selectProps })` - `isFocused` isn't used there and don't exist in argument's props. And therefore TypeScript shows error there because object literals can't have unknown fields. (Actually, I don't understand, how it worked before but anyway). So, I decided to remove `isFocused` in `disabledStyles.control(...)` and `errorStyles.control(...)` since it's not used.
2. What's more important - in `src/components/SelectList/styles.ts` type `CSSObject` picked from `styled-components`, but used widely as `CSSObject` from `emotion`. And their types now are slightly different (`alignContent` is string literal union in emotion, but in SC it's a just string). It's not a problem for functions that return result only for `react-select` (e.g. for `emotion`), but `disabledStyles.label` is used also in `styled-components` context at `src/components/SelectList/components/Label.tsx:28`.
Here, I can't just pass `p` variable to `label(props...)` as `props` variable. So I limited fields: ` label: (props: Pick<SelectListProps, 'variant' | 'size'>): CSSObject => {`. It's also more explicit, IMHO.
3. ❗️ Updated `react-select` now adds `role="combobox"` to `input`. I've updated tests, but I'm not sure, can this broke something in your projects (But it shouldn't - `role="combobox"` fits here well)
4. ❗️ Extending `emotion` component with `styled-components` conflicts at `theme` prop. (`src/components/PhoneInput/PhoneInput.tsx:71`, `src/components/PhoneInput/components/Option.tsx` and `src/components/PhoneInput/components/SingleValue.tsx`). Actually, I don't know how to solve this better, I just suppressed TS errors on that lines for now. I haven't noticed any differences in actual values, it seems that SC overrides the type of source component's `theme` prop but shouldn't.

nlopin (approved):

nlopin:
I'm not sure how can I verify that, YOLO 😄

nlopin:
I've checked on a dummy next.js project and it worked well

PR #225, opened 2022-05-11
feat: add InfoBanner component

### LINKED ISSUE #192: New InfoBanner component

Currently **Wave** does not have implemented the `InfoBanner` component. 
The component purpose as described in Figma:

> Vibrant information messages used as page banners or within forms

 This is a component used by at least a couple of teams and it's quite generic so it would be a great addition to the design system.

[Figma designs](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?node-id=1134%3A24819)

Screenshot 2021-11-11 at 16 44 35

### API
The Props would look something like this.
```typescript
export enum Variants {
    info,
    success,
    warning,
    error,
}
interface InfoBannerProps {
    title: string
    description: string
    variant: Variants
    emphasized: boolean
    linkText?: string
    linkUrl?: string
}
```

martimalek:
**What:**

Adds the `InfoBanner` component following [the designs](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?node-id=357%3A28795).
​
**Why:**

Several projects are using custom implementations of this component, by having it in Wave we will be able to use it as a single source of truth.
​
**Media:**

Playground
Combinations

**Checklist:**

- [x] Base implementation
- [x] Unit tests
- [x] Documentation
- [x] Review the a11y
- [x] Change the text alignment to be left-aligned
- [x] Ready to be merged

Closes #192


src/components/InfoBanner/InfoBanner.tsx
```diff
@@ -0,0 +1,165 @@
...
+        warning: {
+            color: get('semanticColors.icon.primary')
+        },
+        error: {
+            color: get('semanticColors.icon.primaryInverted')
+        }
+    }
+});
+
+export const MinWidthAlertIcon = styled(AlertIcon)`
```
div-Leo, line 108: We're not using this element anymore

src/components/InfoBanner/InfoBanner.tsx
```diff
@@ -0,0 +1,165 @@
...
+    min-width: 1rem;
+`;
+
+const RoundedBox = styled(Box)<BoxWithVariant>`
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-start;
+    border-radius: 0.5rem;
```
JanHamara, line 116: Use **radii[3]** from theme, instead of **0.5rem**?

src/components/InfoBanner/InfoBanner.tsx
```diff
@@ -0,0 +1,165 @@
...
+    border: 0.0625rem solid;
+    padding: 0.5rem 1rem 0.5rem 0.5rem;
```
JanHamara, line 118: Do you think we can use **Spaces** from theme instead of hard coded values here?

src/components/InfoBanner/docs/InfoBanner.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+import { Playground } from 'docz';
+import { InfoBanner } from '../InfoBanner'
+import { InfoBannerPropsTable } from './InfoBannerPropsTable'
+import { StyledSystemLinks } from '../../../docs/StyledSystemLinks'
+
+# InfoBanner
+
+Vibrant information messages used as page banners or within forms.
+
```
JanHamara, line 15: Do you think we could add a little more description and indication of intended usage to Docs?

Something like:

# Usage
Banner is used to communicate a prominent message, state change or important contextual information that affects the system, feature, running process or the whole page. It also can contain related optional actions.

src/components/InfoBanner/InfoBanner.tsx
```diff
@@ -0,0 +1,165 @@
...
+`;
+
+const ICON_VARIANTS = {
+    warning: WarningSolidIcon,
+    info: InfoCircleSolidIcon,
+    success: CheckCircleSolidIcon,
+    error: CloseCircleSolidIcon
+};
+
+export const InfoBanner = ({
```
JanHamara, line 133: Do you think we could call this simply **Banner** ? We don't have a Banner component yet in @wave that would serve a different purpose and I'm not sure if the _Info_ part in name gives more context?

div-Leo (approved):

nlopin (approved):
The implementation is clean 👍 

The only thing that is lacking is an accessibility markup. If we intend to use this banner as an alert, we should follow [the accessibility practices](https://www.w3.org/WAI/ARIA/apg/patterns/alert/) so screen reader and other assistive technologies can notify user that the banner has appeared.

As I understand:
* success and error banners are supposed to have an `alert` role. When they appeared on the page, they will be announced no matter the user's focus
* info and warning banner should have the `status` role + aria-controls (ideally, as a mandatory prop) indicating what is the part of the UI responsible for this banner.

- [ ] Check the a11y of the component

rafael-sepeda (approved):
Thanks!

martimalek:
> The implementation is clean 👍
> 
> The only thing that is lacking is an accessibility markup. If we intend to use this banner as an alert, we should follow [the accessibility practices](https://www.w3.org/WAI/ARIA/apg/patterns/alert/) so screen reader and other assistive technologies can notify user that the banner has appeared.
> 
> As I understand:
> 
> * success and error banners are supposed to have an `alert` role. When they appeared on the page, they will be announced no matter the user's focus
> * info and warning banner should have the `status` role + aria-controls (ideally, as a mandatory prop) indicating what is the part of the UI responsible for this banner.
> 
> * [ ]  Check the a11y of the component

Hey @nlopin, I totally agree with you that the component is currently lacking accessibility, I understand the `alert` and `status` roles based on the banner variant, but I've got a question regarding the `aria-controls`. 
After reading the [MDN docs about aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls), particularly this example:

> An accordion header button toggling the visibility of the accordion panel's content: the accordion header button would have aria-controls set to the ID of the element containing the accordion panel content.

 I understood that the element that "controls" is the one that needs the prop, as I understand it it would be something (e.g. a button, checkbox, etc...) which triggers the InfoBanner to appear and in that case the `aria-controls` should be set in the button/checkbox as the `id` which is set in the InfoBanner, I've never used `aria-controls` so I'm not sure if I'm understanding it properly.

rafael-sepeda:
> Combinations

@martimalek I just realized that in the examples the text is vertically centered.

martimalek:
@rafael-sepeda I didn't notice the alignment issue, I'll fix it before merging

martimalek:
I've fixed the alignment and updated the descriptions image

lloydaf (approved):

src/components/InfoBanner/InfoBanner.tsx
```diff
@@ -0,0 +1,182 @@
...
+    title,
+    description,
+    variant = 'info',
+    linkText,
+    linkUrl,
+    emphasized,
+    ...props
+}: InfoBannerProps): JSX.Element => {
+    const BannerIcon = ICON_VARIANTS[variant];
+    const isInverted = emphasized && variant !== 'warning';
```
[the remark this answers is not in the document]
div-Leo: Do you mean `isInverted`? No, the props are `emphasized` and `variant` then, given these 2 properties, we style the banner.

div-Leo:
@martimalek I was talking to @nlopin and we could start releasing this first version, then we will add the a11y features.

nlopin:
I've created a separate issue to tackle a11y #249

rafael-sepeda:
🔥🔥🔥🔥🔥🔥
@nlopin , @martimalek 
Since we do not have the live demo for this component could you describe what went into implementation so I can tell other designers?

Issue #217, opened 2022-05-02
Change positioning engine library: Migrate from react-tether to react-popper (Proposal)


JanHamara:
# Change positioning engine library: Migrate from [react-tether](https://github.com/danreeves/react-tether) to [react-popper](https://popper.js.org/react-popper/) (Proposal)

This proposal comes up as a result of the integration process of the new [Popover](https://github.com/freenowtech/wave/issues/213) component and realisation of numerous weaknesses and drawbacks of **react-tether** that I believe don't allow the integration of Popover component according to latest best practices and standards in development of React components, as well as according to our Chapter agreements.

A complete comparison in size, usage and features differences between the two libraries (in depth below):

| Library | react-tether | react-popper |
|---|---|---|
| **Core Library** | [tether](https://github.com/shipshapecode/tether) | [popperjs](https://popper.js.org/react-popper/) |
| **Core Lib Latest Release** | 26 March 2021 | 4 days ago |
| **Core Lib Size (minified)** | 21.3kb | 20.5kb |
|---|---|---|
| **React Wrapper Lib** (what we use) | [@react-tether](https://github.com/danreeves/react-tether) | [@react-popper](https://github.com/floating-ui/react-popper) |
| **Latest Release** | 13 Jan **2020** | 4 days ago |
| **Weekly Downloads** | ~ 40 thousand | ~ 6.3 million |
| **Size** | 33 kB (+ 21.3kb @tether) | 5.5 kB (+ 20.5kb @popperjs/core) |
| **Issues / Pull Requests** | 2 / 0 | 12 / 16 |
| **Dependents (projects)** | 172 | 2 338 |
| **Used by (developers)** | ~ 1 200 | ~ 341 000 |
| **React 18 Support** | No | Yes [PR #440](https://www.notion.so/Popper-vs-Tether-cecab5b67e9b4f219adc79be76cc1a2e) |
| **Can use React Portal** | No | Yes |
| **Active Contributors Community** | No | Yes [Discussions](https://github.com/floating-ui/floating-ui/discussions) |

## Maintainability & Currency of the Library
First thing that strikes clear from this comparison is that **react-tether** is not really being developed or maintained anymore, the latest release of core library dates back to over a year ago in March 2021 and the wrapper library even later, latest release having been over 2 years ago - 13 Jan 2020. Meanwhile, **react-popper** seems to keep consistent with releases with latest one 4 days ago.

[...]

PR #230, opened 2022-05-19
feat: Add Popover component to wave

### LINKED ISSUE #213: New component Popover

We need a new component that has the behavior of a Popover. [Here](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/branch/djYaTyPHDwZ8g0MSYWbIcF/Wave-DS?node-id=4119%3A257558) have a draft of the component. It contains also some examples and a possible idea for implementation (to be discussed). 

Currently, in the DS, we have a Tooltip component that is being used wrongly. We want to have a component that opens next to a target by clicking on it and stays open when we move the cursor.

JanHamara:
**What:**

This PR adds a new Popover component to the Wave Design System, as per: [#213]

Using @react-popper lib as per: [#217]
​
**Why:**
Closes #213 
​

**Media:**

There is also Loom presentation of Popover behaviour here -> [#217] 

Screenshot 2022-05-19 at 17 58 57
Screenshot 2022-05-19 at 17 59 09
Screenshot 2022-05-19 at 17 59 24

**Checklist:**
-   [x] Reviewed with designer
-   [x] Ready to merge


src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,155 @@
...
+     * Passes the onClose callback
+     */
+    onClose: (event?: React.KeyboardEvent<HTMLDivElement>) => void;
+}
+
+interface StyledPopoverProps extends React.CSSProperties {
+    isShown: boolean;
+    transform?: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    translate?: any;
```
div-Leo, line 45: What's the `translate` for?

src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+const StyledPopoverParent = styled.div.attrs({ theme })<StyledPopoverProps>`
+    ${({ isShown, transform, top, left, bottom, right, position }) => css`
+        box-sizing: border-box;
+        position: ${position};
+        top: ${top};
+        left: ${left};
+        bottom: ${bottom};
+        right: ${right};
+        transform: ${transform};
```
div-Leo, line 64: can't all of this be passed with the `composed` from the `style-system`. Can we check and, if yes, refactor using those props?

src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,155 @@
...
+}: PopoverContentProps): JSX.Element => {
+    const popoverRef = React.useRef<HTMLDivElement | null>(null);
+
+    const { styles, update } = usePopper(triggerRef, popoverRef.current, {
+        placement,
+        strategy: 'fixed',
+        modifiers: [
+            {
+                name: 'offset',
+                enabled: !!offset,
```
div-Leo, line 100:
```suggestion
                enabled: Boolean(offset)
```
Wdyt?

src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,155 @@
...
+                enabled: true
+            }
+        ]
+    });
+
+    const { popper } = styles;
+
+    React.useEffect(() => {
+        // On each update we need to set focus on the popover content, so that we can close the Popover content with ESC key
+        const timer = setTimeout(() => {
```
div-Leo, line 116: May I ask why the setTimeout?

src/components/Popover/utils.js
```diff
@@ -0,0 +1,105 @@
...
+const transition: Transition = (properties, duration, timingFunction) =>
+    properties.map(property => `${property} ${duration} ${timingFunction}`).join(',');
+
+/**
+ * useStateWithTimeout - Manages state with a Timeout
+ */
+
+type Return<S> = [S, (arg0: S) => void, (arg0: S) => void, () => void];
+
+function useStateWithTimeout<S>(defaultValue: S, timeout?: number): Return<S> {
```
div-Leo, line 18: I can't wrap my head around the need for a timeout? 😬

src/components/Popover/docs/Popover.mdx
```diff
@@ -0,0 +1,113 @@
...
+import { SettingsIcon } from '../../../icons/index';
+
+# Popover
+
+**Primary UI element for content whose visibility can be toggled with a click on trigger element**
+
+Renders a UI component that opens a popup dialog on click of a trigger element.
+<br/>
+
+### Popover Trigger & Content
```
div-Leo, line 23: Can you add the table of props? It should be the `docs/PropsTable`

src/components/Popover/PopoverContent.tsx
JanHamara: These relate to the properties that setup the positioning of Popover content dynamically in `<StyledPopoverParent>`...

This interface and especially the **translate: any** and **transform: string** lines had to be typed here in this fashion, otherwise the build was failing...

Screenshot 2022-05-20 at 13 47 40

I couldn't for life set up types in a way that it wouldn't despite using the type the failing build normally describes for you, in a way such as `Types of property 'translate' are incompatible. Type 'string | number' is not assignable to type ''yes'' | ''no'`' ... 

Setting **translate** as **any** was the only way build was not failing, ( it was failing also with setting it as **unknown** or **never** ), therefore I felt like it was a case of _use **any** only and only if you cannot use anything else_ 🤷

src/components/Popover/PopoverContent.tsx
JanHamara: actually I think this just needed to be **true** by default, as we have a default offset, but good suggestion!

src/components/Popover/PopoverContent.tsx
JanHamara: Because we are using useStateWithTimeout to render the Popover content with a little delay (for the transition animation), therefore we need to also wait that little delay before trying to set focus :)

src/components/Popover/utils.js
JanHamara: -> as said above, opacity transition animation :)

src/components/Popover/utils.js
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * useClickOutside - Handle closing the Popover content when we click outside it
+ */
+
+type UseClickOutside = (
+    ref: { current: HTMLElement | null | undefined },
+    handler: (ev: React.SyntheticEvent<HTMLLinkElement>) => void
+) => void;
+
+const useClickOutside: UseClickOutside = (ref, handler) => {
```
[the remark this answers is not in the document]
div-Leo: I think `transition` should stay, just the hooks.

src/components/Popover/PopoverContent.tsx
div-Leo: Is the animation necessary? I see this logic making the code more prone to bugs and also adding complexity. I'm not sure we generally have animations, so I'm not sure I'd add timeouts to add a barely perceived animation.

nlopin:
@JanHamara can you create an issue to remove TetherJS?

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,240 @@
...
+        ev => {
+            // If openByDefault -> ignore the custom handler
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            if (triggerRef.current && !triggerRef.current.contains(ev.target)) {
+                if (!openByDefault) {
+                    setIsShown(false);
+                    clearIsShownTimeout();
+                    setRenderWithTimeout(false);
+                    resolveCallback(false);
+                } else if (onClose) onClose();
```
nlopin, line 150: let's keep it consistent and avoid oneliners please

src/components/Popover/PopoverContent.tsx
nlopin: Do we understand from where we receive this `yes | no` thing? Maybe we should omit some property

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,240 @@
...
+    &:hover {
+        cursor: pointer;
+        background-color: ${get('semanticColors.background.secondary')} !important;
+    }
+`;
+
+/**
+ * ** Primary UI element for content whose visibility can be toggled with a click on trigger element **
+ *
+ * Renders a UI component that opens a popup dialog on click of a trigger element.
```
nlopin, line 69: Why do we need this duplication? It is available in the mdx

src/utils/hooks/useClickOutside.ts
```diff
@@ -0,0 +1,29 @@
...
+ * useClickOutside - Handle click outside a referenced element (e.g. Popover)
+ */
+type UseClickOutside = (
+    ref: { current: HTMLElement | null | undefined },
+    handler: (ev: React.KeyboardEvent<HTMLDivElement>) => void
+) => void;
+
+const useClickOutside: UseClickOutside = (ref, handler) => {
+    useEffect(() => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
nlopin, line 13: what are those errors? Can we resolve them?

src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,157 @@
...
+        };
+    }, [update]);
+
+    // Close the Popover content with click on ESC key
+    const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+        if (event.keyCode === 27 && onClose) onClose(event);
+    };
+
+    // Hide the Popover content when user clicks outside
+    useClickOutside(popoverRef, event => {
```
nlopin, line 139: have you checked if it works correctly when I click not on the popover ref, but on its children?

For example `ref` is attached to a `div` container with a markup. If I hit the markup (e.g. `<p>` content) the target of the event will be the paragraph, not ref. Hence, you can close the hint when it is not supposed to close.

src/utils/hooks/useHandleKeyDown.ts
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+// Map keys which apply for handleKeyDown
```
nlopin, line 3: a bit obvious 😄

nlopin (commented):
Let's check how it behaves from the accessibility perspective. I think the trigger at least should have `aria-haspopup`, but the exact role needs further investigation. 

Can you take a look at the [WAI ARIA best practices for guidance](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal/)? This is probably not a dialog per se, but something like a popover and I believe there is a role like this in the standard

src/components/Popover/Popover.tsx
JanHamara: These were docs format for Storybook, I will remove it

src/utils/hooks/useClickOutside.ts
JanHamara: this comment has been removed in latest commits

src/components/Popover/PopoverContent.tsx
JanHamara: yes I tested it worked correctly, also this is moved in next commits :)

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,247 @@
...
+    children,
+    content = '',
+    placement = 'bottom-start',
+    offset = 5,
+    isOpen = false,
+    onOpen,
+    onClose
+}: PopoverProps) => {
+    const [triggerReference, setTriggerReference] = React.useState(undefined);
+    const [popperReference, setPopperReference] = React.useState(undefined);
```
martimalek, line 107: Can we call this `contentReference` instead? I find it easier to understand if we use the same terms as in our docs

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    resolveCallback(false);
+                }
+            }
+        },
+        [openByDefault, popoverTriggerRef, handleClose, resolveCallback]
+    );
+
+    const handleKeyControl = (ev: React.KeyboardEvent<HTMLElement>) => {
+        // eslint-disable-next-line default-case
+        switch (ev.keyCode) {
```
martimalek, line 183: What advantage does this switch bring over regular if else? I find it less semantic and we are not using the default case, so for me it looks unnecessary

martimalek, line 183: I personally avoid using switch, I find them unnecessarily verbose and not performant, usually I use an object as a map, in this case it would look like:

```jsx
const KeyToHandler = {
    [KEY_CODE_MAP.ESC]: handleClose,
    [KEY_CODE_MAP.ENTER]: handleClick,
    [KEY_CODE_MAP.SPACE]: handleClick,
}

const handler = KeyToHandler[ev.keyCode]

if (handler) handler()
```

Consider changing it if you like the approach

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,247 @@
...
+            case KEY_CODE_MAP.ENTER:
+                handleClick();
+                break;
+            case KEY_CODE_MAP.SPACE:
+                handleClick();
+        }
+    };
+
+    React.useEffect(() => {
+        if (openByDefault) {
```
martimalek, line 196: Can't we simplify this to `setRender(openByDefault)`?

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,247 @@
...
+            setRender(true);
+        } else {
+            setRender(false);
+        }
+    }, [openByDefault, setRender]);
```
martimalek, line 201: The `setRender` should not be in the dependency array

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,247 @@
...
+     * Optional: Define a callback for when Popover content is opened
+     */
+    onOpen?: () => void;
+    /**
+     * Optional: Define a callback for when Popover content is closed
+     */
+    onClose?: () => void;
+}
+
+export const Popover: React.FC<PopoverProps> = ({
```
martimalek, line 97: Maybe this is a topic to discuss on a different issue, I think we should avoid using `React.FC` if we plan on upgrading React to 18.
Check out [this PR for the breaking changes in the types](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210)

My opinion is that we should ideally only need to type the props, but at the moment we have a linter rule that does not allow that, so maybe for now we can type it as JSX.Element and then remove that rule in a different PR/issue?

src/components/Popover/docs/Popover.mdx
```diff
@@ -0,0 +1,125 @@
...
+import { Link } from '../../Link/Link';
+
+import { Popover } from '../Popover.tsx';
+
+import { PopoverPropsTable } from './PopoverPropsTable.tsx';
+
+
+# Popover
+
+**Primary UI element for content whose visibility can be toggled with a click on trigger element**
```
martimalek, line 22:
```suggestion
**Primary UI element for content whose visibility can be toggled with a click on a trigger element**
```

src/components/Popover/docs/Popover.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+Renders a UI component that opens a popup dialog on click of a trigger element.
```
martimalek, line 24:
```suggestion
Renders a UI component that opens a popup dialog when clicking on a trigger element.
```

src/components/Popover/docs/PopoverPropsTable.tsx
```diff
@@ -0,0 +1,67 @@
...
+                'bottom-end',
+                'left',
+                'left-start',
+                'left-end',
+                'right',
+                'right-start',
+                'right-end'
+            ],
+            description:
+                'Optional: Specify the Popover content placement (placement changes automatically, if Popover content cannot fit inside viewport with selected placement)',
```
martimalek, line 38:
```suggestion
                'Optional: Specify the Popover content placement (it changes automatically if the Popover content cannot fit inside the viewport with the selected placement)',
```

src/utils/hooks/useClickOutside.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
+
+/**
+ * useClickOutside - Handle click outside a referenced element (e.g. Popover)
+ */
+type UseClickOutside = (ref: { current: HTMLElement | null | undefined }, handler: (ev) => void) => void;
+
+const useClickOutside: UseClickOutside = (ref, handler) => {
+    useEffect(() => {
+        const handleClose = event => {
```
martimalek, line 10: Can this `event` be typed?

JanHamara: Unfortunately, I didn't find a way how to type this so that the build wouldn't be failing... 

See the **handleClose** method in which we pass **event** down to **handler** method from second argument - that handler method in Popover component is **handleOut** method that in scope of component deals both with handling closing the Popover content on Keyboard event (using Enter, Space, or ESC keys) as well as with handling closing the Popover content on click outside / MouseEvent.

I spent quite some time trying to type all of those events, but I was getting lost in endless loop of Typescript errors and suggested solutions from TS logs that didn't work. (the _Type 'ABC' is not assignable to type 'DEF'_ kind of suggestions)

**KeyboardEvent**, **MouseEvent**, **SyntheticEvent**, **TouchEvent**, even using **unknown** and **never** would cause the build to fail and suggest other expected interface (that would also fail).

So, I've left either **any** or no type at all, wherever it was necessary.

I could research this more sometime close in the future, so that we can have a guideline for properly typing all of our events like these.

src/components/Popover/Popover.tsx
JanHamara: Yeah, I'm aware of this and of same opinion, but right now we still have a rule requiring it :) Also I agree JSX.Element would be better, however, at this point I think it's maybe better to be consistent and use React.FC same as all other @wave components? (considering we will be removing this from all components eventually)

src/components/Popover/docs/Popover.mdx
JanHamara: I guess, I would then change to ''_**when user clicks on** a trigger element_'', if we want to be more specific 🤔

src/components/Popover/Popover.tsx
martimalek: Sounds reasonable

src/utils/hooks/useClickOutside.ts
martimalek: Ok no worries, sometimes TS is a bit unruly

src/components/Popover/Popover.tsx
JanHamara: this doesn't work tho, have you tested this actually on component, or this was just suggested code?

Also, regarding performance and verbosity, object as a lookup map is not **always** the best option, following are reasons why switch may actually be better here:

1. _Different cases that return same value/code can be mapped together by placing case one after the other._ (which also means I can remove further 2 lines of code, thanks for making me realise that 😆 )
Screenshot 2022-05-30 at 13 57 44

2. We can also control the order of cases for checking. 

3. **Object takes memory**. While using **switch**, or **if...else** does not require any additional memory. -> which we need to consider for in this case, as we won't be triggering this method that often, that it would be, in my opinion, worth it to define object literal for just 2 different options. And for method that will be triggered very scarcely.4. 

4. Does it work faster using Object literal? (taken from article linked below)
Screenshot 2022-05-30 at 14 08 02


I don't wanna write too much, but I would recommend this article: https://medium.com/front-end-weekly/switch-case-if-else-or-a-lookup-map-a-study-case-de1c801d944 :)

martimalek: I didn't test the code, it was just a suggestion, but I find it strange that it does not work.
I understand that using switch for smaller values can have even better performance, but I (personally) don't like the syntax, anyway it was just a suggestion, you can totally leave the switch.

martimalek:
Everything LGTM, I'd just take a closer look at https://github.com/freenowtech/wave/pull/230#pullrequestreview-986356030 before merging

martimalek (approved):

rafael-sepeda (approved):

src/utils/hooks/useStateWithTimeout.ts
```diff
@@ -0,0 +1,36 @@
+import { useState, useRef, useCallback, useEffect } from 'react';
+
+/**
+ * useStateWithTimeout - Manages state with a Timeout
+ */
+type Return<S> = [S, (arg0: S) => void, (arg0: S) => void, () => void];
+
+function useStateWithTimeout<S>(defaultValue: S, timeout?: number): Return<S> {
```
div-Leo, line 8: Let's remove this for now if we're not using it.

Issue #224, opened 2022-05-10
`TableCell` and `TableHeaderCell` don't apply `display` prop although both accept `LayoutProps`


arturmiglio:
-   `@freenow/wave` version: [v1.2.0](https://github.com/freenowtech/wave/releases/tag/v1.20.0)

### Relevant code

```js
<TableCell (...) display='none' />
```

Probably missing `layout` in the `compose` function:

https://github.com/freenowtech/wave/blob/517c13628a2b494702704c570e2327ad81421a84/src/components/Table/components/TableCell.tsx#L26
https://github.com/freenowtech/wave/blob/517c13628a2b494702704c570e2327ad81421a84/src/components/Table/components/TableHeaderCell.tsx#L32

### What was expected to happen?

The prop is applied and the display (at least) can be controlled. The main use-case is responsive tables, where we want to hide certain columns to fit narrower screens. Then we could pass responsive to hide elements in certain breakpoints values like:

```js
<TableCell (...) display={['none', 'table-cell']} />
```

And if `display` prop is finally not allowed, typescript should trigger an error.

### Reproduction

https://codesandbox.io/s/wave-playground-forked-bnn69g?file=/src/App.tsx

lloydaf:
So I dug around a bit, and for those who didn't know (I didn't), the display related props don't [exist](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ec6e5981ab3c59704942f53f976cc0150f2feb39/types/react/index.d.ts#L2457) on tr, td elements. I fixed it as per the suggestion in [this](https://github.com/freenowtech/wave/pull/233/commits/eae2cefe7ad554405b67beb9733399fb45c7e9b0) commit but I am not sure that this is the best approach. 

Additionally, should we add layout prop styling for `TableRow` elements as well? probably not related to responsiveness but wondering if there can be usecase to extend this prop to show/hide a row

arturmiglio:
Moved to the internal backlog.

PR #233, opened 2022-05-30
fix: adding layout to the compose styles

lloydaf:
**What:**

This PR adds `layout` to the `styled-system` compose function, so that layout props can be applied
​
**Why:**

This fixes issue #224 
​
**How:**

​Add the `layout` function to the `compose` function
**Media:**
Watch me [loom](https://www.loom.com/share/496634167f4849cda3ea8ffc4165cf7a)

**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/src/components/Table/components/TableCell.tsx
+++ b/src/components/Table/components/TableCell.tsx
@@ -1,6 +1,6 @@
 import React, { ComponentPropsWithoutRef, FC, useContext } from 'react';
 import styled from 'styled-components';
-import { compose, LayoutProps, textAlign, TextAlignProps } from 'styled-system';
+import { compose, LayoutProps, textAlign, TextAlignProps, layout } from 'styled-system';
 import { TableContext } from '../context/TableContext';
 import { TableProps } from './Table';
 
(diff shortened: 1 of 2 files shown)


lloydaf:
> Do you see other use cases for the `display`? I'm thinking if hiding a column is the only use case, maybe an explicit `hidden` prop will work better here?

As I [commented](https://github.com/freenowtech/wave/issues/224#issuecomment-1141361532) on the issue, I am not sure if `td` elements (or what we want to use instead) should have a configurable `display` property as part of the component API spec.

This implementation follows the suggestions in the issue, and indeed it does solve the problem. Personally I feel that if we don't want to display a row, we should do it declaratively using conditional operators (or even remove it from the row data using a filter 🤷🏻)

As far as thinking about other usecases go, I think that there can be a use where a cell might have use of the `display` property, such as `display:flex`, if it contains more than one item in the cell. 

A good example would be the [monitoring tab in DMT](https://driverservice-crm.prelive.mytaxi.com/office/57/monitoring), where we have multiple items in the `Driver Cancellations` cells.

nlopin:
I'm not sure how the change of the display can affect semantics, but let's start with this solution and see

PR #258, opened 2022-07-07
fix: display semantic colors in table and allow to filter them

martimalek:
**What:**

Fix a bug where the semantic colors where not appearing in the table in the docs.

​
**Why:**

We want to deprecate "regular old colors" in favour of semantic colors, therefore we should offer the best possible experience to our users when working with the latter.

[Check out how the colors do not appear in the table in our docs](https://wave.free-now.com/essentials/colors#functional-reference)

​
**How:**

- Properly converting a `Map` with all of the colors to an `Array` in order to iterate and `.filter` it on the table.

​
**Media:**
Before
image

After
image

**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/src/essentials/Colors/docs/SemanticColorsTable.tsx
+++ b/src/essentials/Colors/docs/SemanticColorsTable.tsx
@@ -32,7 +32,7 @@ export const SemanticColorsTable: FC = () => {
     const [nameSearchInput, setNameSearchInput] = useState('');
     const flatSemanticColors = flattenObj(SemanticColors);
 
-    const filteredColorKeys = Object.keys(flatSemanticColors).filter(it => {
+    const filteredColorKeys = ([...flatSemanticColors.keys()] as string[]).filter(it => {
         if (nameSearchInput === '') {
             return true;
         }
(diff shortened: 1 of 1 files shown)


src/essentials/Colors/docs/SemanticColorsTable.tsx
```diff
@@ -32,7 +32,7 @@ export const SemanticColorsTable: FC = () => {
     const [nameSearchInput, setNameSearchInput] = useState('');
     const flatSemanticColors = flattenObj(SemanticColors);
 
-    const filteredColorKeys = Object.keys(flatSemanticColors).filter(it => {
+    const filteredColorKeys = ([...flatSemanticColors.keys()] as string[]).filter(it => {
```
lloydaf, line 35: One liner alternative: `Array.from(mapObj.keys()).filter(it => !searchStr || it.includes(searchStr))`

martimalek: Makes a lot of sense, I've also moved the `flatSemanticColors` and the `.keys()` outside the component to not have to redeclare them on every render

src/essentials/Colors/docs/SemanticColorsTable.tsx
```diff
@@ -28,17 +28,15 @@ const ColorBlock = styled.div<{ color: string }>`
...
-    const filteredColorKeys = Object.keys(flatSemanticColors).filter(it => {
-        if (nameSearchInput === '') {
-            return true;
-        }
 
-        return it.toLowerCase().includes(nameSearchInput.toLowerCase());
-    });
+    const filteredColorKeys = !nameSearchInput
+        ? flatSemanticColorsKeys
+        : flatSemanticColorsKeys.filter(it => it.toLowerCase().includes(nameSearchInput.toLowerCase()));
```
nlopin, line 39: Let's remove spaces too
```suggestion
        : flatSemanticColorsKeys.filter(it => it.toLowerCase().includes(nameSearchInput.toLowerCase().trim()));
```

nlopin (approved):

src/essentials/Colors/docs/SemanticColorsTable.tsx
martimalek: Done

Issue #259, opened 2022-07-08
Expose event props onFocus and onOpen from the PhoneInput


nlopin:
We want to track several user actions performed on the PhoneInput, but it is impossible with the current API.

* send an event when prefix select is opened
* send an event when the input field is focused

As a solution, I think we should expose three props to stay consistent:
* `onPrefixFocus` - when prefix select is focused
* `onPrefixOpen` - when the prefix select is open
* `onInputFocus`

nlopin:
The agreed API is:

```
<PhoneInput
        menuPortalTarget={document.body}
        inputProps={{
          onFocus: ()=> {}
        }}
        selectListProps={{
          onFocus: () => {
        }}}
      />
```

Where everything from the `inputProps` is passed to the phone input and everything from the `selectListProps` to the prefix list

PR #261, opened 2022-07-14
Expand PhoneInput API

### LINKED ISSUE #256: PhoneInput view cut
-   `@freenow/wave` version: 1.25.1

### Relevant code

```js
var your => (code) => here;
```

After update of the Wave PhoneInput drop down displays cut (currently is not possible to solve via menuPortalTarget as this prop is not being passed)

### Reproduction

Try to register a new driver on https://registration.prelive.free-now.com/ (e.g. Paris - Taxi) on contacts detail page PhoneInput is used.

Screenshot 2022-07-06 at 13 27 00

nlopin:
## What

Allows to attach portals and exposes implementation level APIs
​
**Why:**

Closes #256 #259

​
**How:**

- adds `listPortalTarget` prop to be able to attach the prefixes popup to another HTML element
- adds `inputProps` and `selectListProps` props and passes them to the corresponding components. Those props has less priority than higher level props of the `PhoneInput` (e.g. `<PhoneInput id='hello' inputProps={{id: 'input'}} />` will use _hello_ as the id)
- updated the documentation
​
**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/src/components/PhoneInput/docs/PhoneInput.mdx
+++ b/src/components/PhoneInput/docs/PhoneInput.mdx
@@ -53,7 +53,7 @@ return (
 
 ## Troubleshooting
 
-### The list of previxes is not fully visible
+### The list of prefixes is not fully visible
 
 If the list of prefixes is not fully visible when opened, like in [this Sandbox example](https://codesandbox.io/s/inspiring-austin-2006xl?file=/src/index.js), use `listPortalTarget` prop to attach the popup to the document's body:
 
--- a/src/components/PhoneInput/docs/PhoneInput.mdx
+++ b/src/components/PhoneInput/docs/PhoneInput.mdx
@@ -15,6 +15,10 @@ import { COUNTRIES } from '../constants';
 
 # PhoneInput
 
+The `PhoneInput` is a component to input a phone number in international format.
+
+The component consists of two controls: a select to pick a prefix and an input to type in the phone number. The select shows all the country codes available world-wide.
+
 ## Properties
 
 <PhoneInputPropsTable />
@@ -40,3 +44,9 @@ return (
     />
 );
 ```
+
+## Playground
+
+<Playground>
+    <ControlledPhoneInput label="Mobile Phone Number" />
+</Playground>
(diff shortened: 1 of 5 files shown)


div-Leo (approved):

src/components/PhoneInput/docs/PhoneInput.mdx
```diff
@@ -40,3 +44,21 @@ return (
...
+
+### The list of prefixes is not fully visible
+
+If the list of prefixes is not fully visible when opened, like in [this Sandbox example](https://codesandbox.io/s/inspiring-austin-2006xl?file=/src/index.js), use `listPortalTarget` prop to attach the popup to the document's body:
+
+```jsx
+<PhoneInput listPortalTarget={document.body} />
+```
+
+The reason of the issue is that the prefix selection popup is rendered inside the same container as the component by default. If the container do not allow content to overflow (e.g. via `overflow: hidden;`), everything that overflows the container borders will be hidden.
```
martimalek, line 64:
```suggestion
The reason of the issue is that the prefix selection popup is rendered inside the same container as the component by default. If the container does not allow content to overflow (e.g. via `overflow: hidden;`), everything that overflows the container borders will be hidden.
```

src/components/PhoneInput/docs/PhoneInputPropsTable.tsx
```diff
@@ -56,6 +56,25 @@ export const PhoneInputPropsTable: FC = () => {
...
+            description:
+                'Pass props directly to the internal input component. ' +
+                'Any value from the `Input` component props are allowed, but props from the `PhoneInput` take precedence'
+        },
+        {
+            name: 'selectListProps',
+            type: 'Object',
+            description:
+                'Pass props directly to the internal SelectList component used to show prefixes. ' +
+                'Any value from the `Input` component props are allowed, but props from the `PhoneInput` take precedence'
```
martimalek, line 77:
```suggestion
                'Any value from the `SelectList` component props are allowed, but props from the `PhoneInput` take precedence'
```

martimalek (approved):
Great job!

Issue #232, opened 2022-05-27
Remove react-tether positioning library


JanHamara:
### Describe the feature you'd like
Based on RFC #217 - we have decided to replace **react-tether** positioning engine with **react-popper**

### Describe alternatives you've considered

**@floating-ui/react-dom** - Discussion also on #217

### What changes need to be done
-   [x] Refactor Tooltip component to use **react-popper**
-   [ ] Refactor Datepicker component to use **react-popper** (possibly reuse Popover component for Datepicker)
-   [ ] Remove **react-tether** dependency from the project

PR #267, opened 2022-07-28
feat: Tooltip with react-popper

dornelasN:
**What:**
Replace **react-tether** positioning engine with **react-popper**
​
**Why:**
Based on the issue #232  and RFC #217, we are replacing the usage of react-tether for **react-popper**
​
**How:**
We created a small map function to map the `TooltipPlacement` values into the react-popper's `Placement` values to use in the `usePopper` hook of `react-popper`.

Using the `styles` and `attributes` from the `usePopper` hook  we can set the references for both the tooltip trigger and content to then set its placement and styles.

With the `usePopper` hook we were able to remove the global styles previously created to override the tether's by also using the attributes returned from the hook to set the Tooltips` arrow placement with the `styled-system's variant`.
​
**Media:**

It looks the same as before, so media doesn't seem necessary.

**Checklist:**

-   [x] Ready to be merged


src/components/Tooltip/Tooltip.tsx
```diff
@@ -152,6 +120,23 @@ const Tooltip: React.FC<TooltipProps> = ({
     inverted = false
 }: PropsWithChildren<TooltipProps>) => {
     const [isVisible, setIsVisible] = React.useState(alwaysVisible);
+    const [triggerReference, setTriggerReference] = React.useState(undefined);
```
lloydaf, line 123: Can you add some comments here explaining the purpose of the state variables and the mapping function at a high level? I don't assume people needing to know this unless they need to change something, but good to have

src/components/Tooltip/Tooltip.tsx
```diff
@@ -171,30 +156,22 @@ const Tooltip: React.FC<TooltipProps> = ({
...
+                onMouseOver: () => handleVisibilityChange(true),
+                onMouseOut: () => handleVisibilityChange(false),
+                ref: setTriggerReference
+            })}
+            {isVisible && (
+                <TooltipBody
+                    ref={setContentReference}
+                    inverted={inverted}
+                    style={{ ...styles.popper }}
+                    variant={attributes.popper?.['data-popper-placement']}
```
lloydaf, line 169: Is this the recommended way of doing this?

src/components/Tooltip/TooltipPlacement.ts
```diff
@@ -7,3 +9,17 @@ export type TooltipPlacement =
     | 'top-right'
     | 'center-left'
     | 'center-right';
+
+const tooltipPlacementMap = {
+    'bottom-left': 'bottom-start',
```
lloydaf, line 14: Similar to how types have been defined for `TooltipPlacement`, can we also define them for `PopperPlacement` _(is that the same as `Placement`?)_ and take advantage of that to type `tooltipPlacementMap`?

src/components/Tooltip/TooltipPlacement.ts
```diff
@@ -7,3 +9,17 @@ export type TooltipPlacement =
...
+    'bottom-center': 'bottom',
+    'bottom-right': 'bottom-end',
+    'top-left': 'top-start',
+    'top-center': 'top',
+    'top-right': 'top-end',
+    'center-left': 'left',
+    'center-right': 'right'
+};
+
+export const mapTooltipPlacementToPopperPlacement = (tooltipPlacement: TooltipPlacement | Placement): Placement =>
```
lloydaf, line 24: Do you see an advantage to having this function vs. accessing it from the map object directly?

src/components/Tooltip/TooltipPlacement.ts
JanHamara: I am not sure why are we trying to keep original placement values and try to map them to @popperjs Placement type values, when we could just directly use Placement values from @popperjs/core/lib/enums?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -45,83 +88,8 @@ const TooltipBody = styled.div<Pick<TooltipProps, 'inverted'>>`
         border: 0.25rem solid rgba(0, 0, 0, 0);
         border-bottom-color: ${p => (p.inverted ? Colors.AUTHENTIC_BLUE_50 : Colors.AUTHENTIC_BLUE_900)};
         margin-left: -0.25rem;
```
JanHamara, line 90: what is this negative left margin used for?

JanHamara (changes requested):

src/components/Tooltip/Tooltip.tsx
div-Leo: It's a requirement for usePopper to work. I don't understand the concern. Can you explain better what you think is the issue?

src/components/Tooltip/TooltipPlacement.ts
div-Leo: We are keeping them because otherwise we will introduce breaking changes. The idea is to communicate that those value are deprecated, but we can't remove them now.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -152,6 +120,23 @@ const Tooltip: React.FC<TooltipProps> = ({
...
 }: PropsWithChildren<TooltipProps>) => {
     const [isVisible, setIsVisible] = React.useState(alwaysVisible);
+    const [triggerReference, setTriggerReference] = React.useState(undefined);
+    const [contentReference, setContentReference] = React.useState(undefined);
+
+    const mappedPlacement = mapTooltipPlacementToPopperPlacement(placement) ?? (placement as Placement);
+
+    const { styles, attributes } = usePopper(triggerReference, contentReference, {
+        placement: mappedPlacement,
+        modifiers: [
```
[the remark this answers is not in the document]
div-Leo: We tried and didn't see any difference would you mind exerting on this? Why is it necessary and what is defference

src/components/Tooltip/Tooltip.tsx
```diff
@@ -152,6 +120,23 @@ const Tooltip: React.FC<TooltipProps> = ({
...
+            {
+                name: 'offset',
+                enabled: true,
+                options: {
+                    offset: [0, 5]
+                }
+            }
```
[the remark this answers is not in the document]
div-Leo: Not necessary to scroll as it's a tooltip on scroll it will go away and re render in the correct position

src/components/Tooltip/Tooltip.tsx
div-Leo: The arrow

src/components/Tooltip/Tooltip.tsx
div-Leo: This is to render the arrow dynamically. Is the only option we found... We saw a library that is called `react-popper-tooltip` and check its implementation and it was using this key to get the actual current placement as the one we get from the props is not reliable because of popper which changes the placement based on the screen size and screen positioning of the tooltip.

About the `popoverTriggerRef` I don't think you're actually using it in the `Popover`, it's being passed to a component but not really used. The used one are the `triggerReference` and `contentReference`. 

> Also what is this variant trying to do with attributes.popper? set up the positioning of the arrow?
Yes

> I wonder if it wouldn't have been easier to just copy the Popover implementation
It's what we did but there are many things that are different and not necessary.

src/components/Tooltip/TooltipPlacement.ts
div-Leo: @dornelasN Let's check this

src/docs/PropsTable.tsx
```diff
@@ -32,7 +32,7 @@ export const PropsTable: React.FC<PropsTableProps> = ({ props: componentProps }:
                                 {prop.name}
                             </Text>
                         </TableCell>
-                        <TableCell>
+                        <TableCell style={{ whiteSpace: 'normal' }}>
```
dornelasN, line 35: We wanted to break the line inside of the `PropsTable` inside the documentation as we have many different options for the `Placement` of a Tooltip, adding the whiteSpace with either `normal` or `break-spaces` seems to do the deal with the following:

image

Initially Leo and I thought about having a parameter to send to the PropsTable to define this wrapping, but I've checked all other tables and it does not seem to affect them as this might be the first one that we have so many possible options. So I'm adding this directly to the Type's `TableCell` if that's okay. If you have any other recommendation to achieve this, let me know please :)

src/components/Tooltip/Tooltip.tsx
```diff
@@ -18,7 +19,75 @@ const fadeAnimation = keyframes`
...
+            transform: 'rotate(-90deg)'
+        },
+        // TODO: Fix arrow placement
+        'right-end': {
+            bottom: '0.5rem',
+            left: '-0.25rem',
+            right: 'auto',
+            transform: 'rotate(-90deg)'
+        },
+        // TODO: Fix arrow placement
```
nlopin, line 75: What's the issue here? I don't think we can publish a component with broken arrows

src/components/Tooltip/Tooltip.tsx
```diff
@@ -147,11 +141,37 @@ interface TooltipProps {
...
+    const [triggerReference, setTriggerReference] = React.useState(undefined);
+    const [contentReference, setContentReference] = React.useState(undefined);
+
+
+    /**
+     * Map the older placement values to Popper placement  as we need to get the correct placement for the tooltip from the Popper library
+     * without introduce any breaking changes to the Tooltip component.
+     * TODO: Remove in the next major release.
+     */
+    const mappedPlacement = mapPlacementWithDeprecationWarning(placement);
```
nlopin, line 161: 👍

nlopin (approved):

src/components/Tooltip/Tooltip.tsx
dornelasN: wow, nelson strikes again. forgot to remove the comments.

div-Leo (approved):

martimalek (approved):

PR #275, opened 2022-08-02
docs(props-table): add props table to Table docs

### LINKED ISSUE #272: Add Table Props Table
We disabled the auto generated documentation for component props because it was too slow. Now we're missing said documentation so we have to add it manually. We have already created a new component to handle displaying the props and it's in use for all new components already. Some old components need to be updated to use the new prop table (as seen in the Button component for example)

The following components need to be updated to use the new props documentation:

- [ ] Table (component)

arturmiglio:
**What:**

Add Properties table for the Table components.
​
**Why:**

Closes #272 

**Media:**

localhost_3001_components_Table

**Checklist:**

-   [ X ] Ready to be merged


nlopin:
Great work! 

I see some components also have props from styled system (MarginProps, HeightProps, WidthProps). We have a component `StyledSystemLinks` to show a sentence with links below the table.

Could you add those where applicable?

`<StyledSystemLinks component="Table" supportedProps={ ['margin', 'width', 'height'] }/>`

You can find an example in the [Button.mdx](https://github.com/freenowtech/wave/blob/main/src/components/Button/docs/Button.mdx?plain=1#L16)

martimalek (commented):
Awesome! Thanks a lot 🙇🏽

arturmiglio:
> I see some components also have props from styled system (MarginProps, HeightProps, WidthProps). We have a component StyledSystemLinks to show a sentence with links below the table.

Since I didn't know about the StyledSystemLinks, I've manually 'parsed' what would make sense to have on the table. Do you think we can keep both? Or drop the table and leave only the links instead?

With tables and links:

Screenshot 2022-08-03 at 12 47 02

Without:

Screenshot 2022-08-03 at 12 51 31

martimalek:
@arturmiglio I think that so far when one of our components supports styled-system props we only use the `StyledSystemLinks`, so I'd drop the table for those cases to be consistent. Also this way seems less redundant and more clean imo

arturmiglio:
> @arturmiglio I think that so far when one of our components supports styled-system props we only use the `StyledSystemLinks`, so I'd drop the table for those cases to be consistent. Also this way seems less redundant and more clean imo

✅ Done!

martimalek (approved):

nlopin (approved):

lloydaf:
merging to feature branch

PR #276, opened 2022-08-03
Issue 10 add prop documentation

### LINKED ISSUE #10: Add missing component prop documentation for components
We disabled the auto generated documentation for component props because it was too slow. Now we're missing said documentation so we have to add it manually. We have already created a new component to handle displaying the props and it's in use for all new components already. Some old components need to be updated to use the new prop table (as seen in the Button component for example)

The following components need to be updated to use the new props documentation:

- [x] Checkbox
- [x] Datepicker
- [x] Filepicker
- [x] Headline
- [x] HelperText
- [x] Link
- [x] Pagination
- [x] RadioButton
- [x] Select
- [x] TabBar
- [x] #272
- [x] Tag
- [x] Text
- [x] Textarea
- [x] Toggle
- [x] Tooltip

lloydaf:
**What:**

This PR adds documentation of the props of components that didn't have it defined before. We are using the `PropsTable` component for the same
​
**Why:**

Closes #10 
​
**How:**

1. Create a `<component>PropTable.tsx` file to define the props according to the `PropsTable` API
2. Import it in the component doc file
​
**Media:**
The documentation for `Textarea` looks as shown below:

Screenshot 2022-08-03 at 09 06 47

**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @nlopin @martimalek @div-Leo @rafael-sepeda @jonotrujillo
+* @nlopin @martimalek @div-Leo @jonotrujillo
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Thanks for being willing to contribute! Please read the [CONTRIBUTING.md](./CONT
 
 ## Have doubts?
 
-Please don't hesitate to let know us what are your doubts, what can be improved or what is difficult for you to grasp from the documentation. Open issues to start the conversation! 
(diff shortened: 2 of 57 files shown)


martimalek:
I see that some files appear as changes introduced in this PR when they should already be in master, is this up to date? Could you double check?

lloydaf:
@martimalek the issue is that the default merge strategy is to squash. So the commit histories of both look different, that's why they show up on the PR. As you mentioned those changes are already on master

nlopin (approved):
Looks great 👍

PR #281, opened 2022-08-18
fix: conflicts between next and main v2

martimalek:
**What:**

The past month I opened a PR where I tried to resolve conflicts between `main` and `next` (#257). That didn't really solve the issue, the conflicts still appeared. This PR aims to solve this issue once and for all.

​
**Why:**

It's making our CI fail.

​
**How:**

I've created a branch from `next`, merged `main` into it and accepted all incoming changes.

-   [x] Ready to be merged


.github/workflows/library.yml
```diff
@@ -29,7 +29,7 @@ jobs:
               run: npm run test
 
     release:
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/next'
```
nlopin, line 32: I don't think we want to release next version. Maybe under `-next` for internal testing. What was your idea here?

package.json
```diff
@@ -88,6 +90,7 @@
         "jest-axe": "^3.4.0",
         "jest-date-mock": "^1.0.8",
         "jest-styled-components": "^7.0.2",
+        "jscodeshift": "^0.13.0",
```
nlopin, line 93: 😍

docs/migrating.mdx
```diff
@@ -0,0 +1,30 @@
...
+A codemod exists to upgrade your components for this version.
+
+```bash
+npx jscodeshift -t node_modules/@freenow/wave/codemods/v2.ts path/to/src
+```
+
+### Button and TextButton
+
+The `block` property used to act as a shortcut to give the button components a width of 100%. In the future, use the `width` property
+directly. It uses the styled-system variable, which is a lot more flexible than just the boolean flag.
```
nlopin, line 19: What do you think about different phrasing:

```suggestion
The `block` property is removed. Use `width='100%'` instead. The `width` allows to set value for each breakpoint which is a lot more flexible than just the boolean flag.
```

nlopin (commented):
Great initiative Marti

nlopin (approved):

Issue #203, opened 2022-01-28
Migrate all components to semantic colors


nlopin:
Part of #121 initiative

Replace all `Color` calls to `get("semantic_color_path")`

PR #284, opened 2022-09-08
refactor: migrate first 5 components to SemanticColors

martimalek:
**What:**

Start migrating components to use semantic colors.

​
**Why:**

Part of #203 

​
**How:**

Use `SemanticColors` instead of `Colors`in:
- Accordion
- Banner
- Box
- Button
- Toggle
- Tooltip

- [x] Ready to be merged
- [x] Need design input in one TODO

### DIFF
--- a/src/components/Toggle/Slide.tsx
+++ b/src/components/Toggle/Slide.tsx
@@ -1,16 +1,16 @@
 import styled from 'styled-components';
-import { Colors } from '../../essentials';
+import { SemanticColors } from '../../essentials';
 
 const determineBackground = (props: SlideProps) => {
     if (props.disabled) {
-        return Colors.AUTHENTIC_BLUE_50;
+        return SemanticColors.background.secondary;
     }
 
     if (props.error) {
-        return Colors.NEGATIVE_ORANGE_900;
+        return SemanticColors.background.dangerEmphasized;
     }
 
-    return Colors.ACTION_BLUE_900;
+    return SemanticColors.background.infoEmphasized;
 };
 
 interface SlideProps {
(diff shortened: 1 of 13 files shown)


src/components/Toggle/Slide.tsx
```diff
@@ -23,7 +23,10 @@ const Slide = styled.div<SlideProps>`
     height: 1rem;
 
     cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
-    background: ${props => (props.disabled ? Colors.AUTHENTIC_BLUE_50 : Colors.AUTHENTIC_BLUE_200)};
+    background: ${props =>
+        props.disabled
+            ? SemanticColors.background.secondary
+            : SemanticColors.border.primary}; // TODO should this come from background
```
martimalek, line 29: I think that it would make more sense for this to be `SemanticColors.background.{something}`, but we don't have this color (AUTHENTIC_BLUE_200) there at the moment

nlopin: Can you sync with someone in the design team? Pietro, Vini or Fernanda?

nlopin (approved):
All good. Semantic naming needs some naming convention because sometimes the name contains states encoded (e.g. `linkHover`), but clarification messages follow the same pattern (e.g. `secondaryEmphasized`).

It is definitely out of the scope of the PR, so LGTM

src/components/Toggle/Slide.tsx
martimalek: Hey @nlopin, sorry for taking so long, I checked the [Figma file](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?node-id=325%3A2) for the Toggle and I see that the colors (with semantic names) are included but we are not following the same convention in the code.
In the Figma there is a `colors.forms.toggle.default.frameOff` but we don't even have the `forms` category.

image

Should we be aligned so that the tokens in Figma and our colors structure are the same? I believe so but I want to double check before starting a conversation with designers about this.

nlopin: I think we should. Let's find who can we sync with. Ideally we can ask designers to export their tokens from Figma and use them

martimalek: I've asked them about it, [their colors](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?node-id=137%3A0) already look a lot like ours, so I think it's just a matter of adding the ones that we are missing in the code. I've created #286 to take care of that.

PR #288, opened 2022-10-04
fix(check-icon): remove fill none

arturmiglio:
**What:**

Fix the colour of the CheckIcon.
​
**Why:**

The CheckIcon svg tag has a fill none rule that makes it transparent in some situations.

​
**How:**

Remove `fill='none'` from the svg tag.

​
**Media:**

Screenshot 2022-10-04 at 12 08 18


martimalek:
@arturmiglio What is the status on this? Do we need anything else?

arturmiglio:
@div-Leo @JanHamara I've just had a chat with some PDs about this. More info on how will we proceed soon, after they have some time to analyse the issue and come up with a solution for the Icons exporting flow (let's call it like this, shall we?).

I'll keep you posted. Thanks!

nlopin:
@artur-miglio-free-now any update on it? If not, we'll close the PR in two weeks

arturmiglio:
> @artur-miglio-free-now any update on it? If not, we'll close the PR in two weeks

Thanks for following up on this. I'll bring the topic back up with the PDs.

arturmiglio:
I'll close it for now and re-open if that's the case. Please don't delete the branch though.

PR #292, opened 2022-10-25
fix: pass inherit to the entire svg

div-Leo:
**What:**
Some icons do not inherit the color when inside the button
​
**Why:**
We set a rule to inherit the color just to `svg path` and was not including some SVGs elements
​
**How:**
Changed the class to include everything inside the SVG (e.g. `g`, `circle`, `use`)
​
**Media:**
Screenshot 2022-10-25 at 18 02 10

If approved reject #288


div-Leo:
Closing because we're going to follow up on #288 with a different strategy.

PR #300, opened 2023-02-10
Bugfix/build fix

div-Leo:
**What:**
Fix docz build

**Why:**
It was not running with the latest versions of node.
​
**How:**

- Added `NODE_OPTIONS=--openssl-legacy-provider` flag 
- recompiled `package-lock.json`
- cleaned the .docz cache​


jonotrujillo (approved):
Thanks for fixing it!

martimalek (approved):

src/components/Table/components/Table.tsx
```diff
@@ -41,6 +46,7 @@ const Table: FC<TableProps> = ({
     };
 
     return (
+        // @ts-expect-error table props (height, width) conflict with HeightProps and WidthProps from styled-system
```
nlopin, line 49: I think it should be `ts-ignore`

lloydaf: should we update TS to 3.9 instead?

lloydaf: https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#ts-expect-error-comments

nlopin (approved):

src/components/Table/components/Table.tsx
nlopin: I'll do it in a follow-up

Issue #240, opened 2022-06-03
Migrate to `useId` from `nanoid` 


nlopin:
The new React Engine (#239) exposes new [`useId` hook](https://reactjs.org/docs/hooks-reference.html#useid) which can generate stable ids both on the server and frontend side.

That means we can migrate to this hook and remove `nanoid` from the project which solved the same task (not successfully, because ids were not the same on BE and FE sides)

martimalek:
Closing it since it was tackled in #321 and will be merged to `next` with #311

Issue #239, opened 2022-06-03
Switch to React Concurrent Engine (stop supporting react<18)


nlopin:
As a result of the discussion #212

The new major should rely on the new React Engine. That means we need to change the minimum required version of react in `package.json`.

This also means we can use new React@18 hooks in the library

lloydaf:
We have a lot of tech debt that we need to clean up in BOMT (and maybe even DMT) before we can migrate to react18. I think it's really nice to be able for wave to have react18 support, but dropping support for previous versions is concerning imo, because it wouldn't let older legacy projects to receive updates. Correct me if I misunderstood
@div-Leo @JanHamara do you have any thoughts?

nlopin:
I think we need to create some pressure from the library side of things to move things (DMT and BOMT) forward. It will be easier to justify the refactoring

nlopin:
@lloydaf and @martimalek researched the topic. 

They've changed Wave's required react version to 18 and used a new `useId` hook in the `useGeneratedId` hook. Many essential components use this hook (Input, Textarea, Datepicker, etc.).

After that, they took the oldest project that we have (the DMT) and tried to use this react-18-only library in several scenarios:

- The DMT uses the version of React that we're using now (16.8). The app crashed ❌
- The DMT was promoted to React 18 without switching to the new engine. The app worked with a warning recommending switching to the new engine ✅
- The DMT was promoted to React 18 with the new engine. All good ✅

The learning here is that we can support react 18 only with the new release. This will be a good driver for our applications to migrate to the latest version of react. The migration itself is not complicated, but the migrated app needs validation (manual or e2e)

More info in the [internal issue](https://jira.intapps.it/browse/WECH-21)

PR #310, opened 2023-03-13
chore: migrate to react 18

martimalek:
**What:**

Migrate to react 18

​
**Why:**

We want to stop supporting react<18 in Wave 2.0 (#239)
​
**How:**

- Bump react and react-dom versions to ^18.2.0 (not sure if we should target ^18.0.0 instead)
- Use react's `useId` instead of the internal `generateId` in `useGeneratedId`

**Checklist:**

- [ ] Fix type errors
- [ ] Use react's `useId` in `Month.tsx` instead of the internal `generateId`
- [ ] Remove our internal util `generateId` once it's no longer used
- [ ] Remove `nanoid` dependency
- [ ] Remove IE11 polyfill required for `nanoid` once it's no longer used (basically remove `polyfills.ts`)
- [ ] Fix dependency errors (`docz` is not compatible with React 18)
- [ ] Ready to be merged

Since we already refactored `useGeneratedId` to not use `nanoid` I think it makes sense to finish removing `nanoid` in this PR, that would mean that #240 is a duplicate, I'm keeping it alive for now to hear other's opinions


martimalek:
I'm closing this since I was using a fork and now I've opened the PR directly from the branch of this repo in #311

PR #316, opened 2023-03-29
Wech 23/migrate components 3

aspasov86:
Migrate the following components from Docz to Storybook

Input
Label
Link
Modal
Offcanvas


src/components/Input/docs/Input.stories.mdx
```diff
@@ -0,0 +1,168 @@
...
+            <button type="submit">Submit</button>
+        </form>
+    );
+
+    const usernameInput = screen.getByRole('textbox', { name: 'username' });
+    const passwordInput = screen.getByLabelText(/password/i);
+    const submitBtn = screen.getByRole('button', { name: /submit/i });
+
+    userEvent.type(usernameInput, 'jaimito');
+    userEvent.type(passwordInput, 'tontoelquelolea');
```
martimalek, line 107: I know you didn't introduce this now, but maybe it would be nice to change this? 🤣 🤣

martimalek (approved):

martimalek (commented):
Shouldn't we be removing the old `.mdx` files?

src/components/Input/docs/Input.stories.mdx
```diff
@@ -0,0 +1,168 @@
...
+    argTypes={{
+        variant: {
+            options: ['boxed', 'bottom-lined'],
+            defaultValue: 'boxed',
+            control: 'select'
+        },
+        size: {
+            options: ['small', 'medium'],
+            defaultValue: 'medium',
+            control: 'select'
```
JanHamara, line 21: I think it would be good idea to aim to cover all props in `argTypes` with this migration (inverted, placeholder, etc.) with relevant controls per each prop type, so that our components are fully interactive on the Canvas tab (especially as this becomes a single view in Storybook v7, while Docs tab will be moved to sidebar)

nlopin: They are derived automatically by the Storybook, aren't they?

JanHamara: They may be derived for documentation purposes yes, but in order to interact with them in Canvas (which I think is probably the most important feature of SB), we need to declare their controls like here in `argTypes` (unless that changed in SB v7)

PR #317, opened 2023-03-29
Wech 23/migrate components 4

christify30:
**What:**
converting  table,tag,text,textarea,toggle and tooltip docs to stories

​
**Why:**
The previous documentation expired

​
**How:**
Create a relevant story file

​
**Media:**

**Checklist:**

-   [ ] Ready to be merged


package-lock.json
martimalek, line 1: I wouldn't add this since it's not related to the PR

src/components/Table/docs/Table.stories.mdx
```diff
@@ -0,0 +1,130 @@
+import { LiveTable } from './LiveTable';
```
martimalek, line 1: Shouldn't each component have at least one `Story`? I'm not very familiar with Storybook and I'm not sure what the canvas will show f we don't have any

src/components/Table/docs/Table.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+import { ActiveRowTable } from './ActiveRowTable';
+import { ComplexDataTable } from './ComplexDataTable';
+import { SkeletonTable } from './SkeletonTable';
+import { TablePropsTable, TableRowPropsTable, TableRowSkeletonPropsTable } from './TablePropsTable';
+import { StyledSystemLinks } from '../../../docs/StyledSystemLinks';
+
+# Table
+
+<Meta title="Components/Table" component={Table} />
```
JanHamara, line 10: I think it would be good idea to aim to cover all props in argTypes with this migration (rowStyle, rowSize, columnSpace) with relevant controls per each prop type, so that our components are fully interactive on the Canvas tab (especially as this becomes a single view in Storybook v7, while Docs tab will be moved to sidebar)

This would apply to all migrated components :)

src/components/Table/docs/Table.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+
+export const Template = args => <Table {...args} />;
+
+Tables are used to render tabular data in a basic grid, making it easier for people to scan it. Tables are comprised of cells, columns, and rows.
+
+## Properties
+
+### Table
+
+<TablePropsTable />
```
JanHamara, line 20: With features that Storybook provides us with, I believe we don't need our custom props tables anymore, Storybook generates one for us with all information if we describe/comment our props interfaces correctly, define argTypes, etc.

src/components/Toggle/docs/Toggle.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+-   Use toggle to swap the state of a functionality or preference (usually between „on“ and „off“)
+-   For multi-state scenarios consider radio buttons or checkboxes
+-   Toggles should take affect automatically (without saving)
+
+## Technical Information
+
+Inside the toggle component is a checkbox and therefore the api and usage works similarly. For example, instead
+of passing the `value` prop, the `checked` prop must be used instead.
+
+## Playground
```
JanHamara, line 39: Can we name this **Stories** as `<Canvas>` and `<Story>` components don't exactly provide a Playground for us, as before, that's replaced with Canvas tab now :)

src/components/Textarea/docs/Textarea.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+    >
+        {Template.bind({})}
+    </Story>
+</Canvas>
+
+## Appearance
+
+### Default
+
+<Textarea label="Default" />
```
JanHamara, line 40: Can we take the Storybook approach here and use `<Canvas>` and `<Story>` components?

Example: src/components/Password/docs/Password.stories.mdx in https://github.com/freenowtech/wave/pull/319/files#diff-d446ba0b705dc09e92956a803cb2dfd54917947ccea20dfa0635c3280ef0a336

src/components/Table/docs/Table.stories.mdx
JanHamara: Yeah, you are correct, we should declare at least the most important test cases with `<Canvas>` and `<Story>` components. e.g. here we could have at very least 3 stories for each table row style. 

I'd also probably add story for table with active row, loading state, sortable table.

nlopin: There are interactive examples on the page, I'm not sure how we can make stories out of them because they have to contain internal state

Issue #169, opened 2021-08-16
Make `ref`s point to "correct" elements


nlopin:
The outcome of the dialog #144

Example code: https://codesandbox.io/s/wave-input-refs-c5ure

Currently, the `ref` is attached to the wrapper which contradicts [our agreement](https://github.com/freenowtech/wave/issues/156#issuecomment-880546292) to attach to the most obvious element.  We need to attach the ref to the `input` and research for other use cases needed to be fixed.

Components using `forwardRef` that need correction:
- [ ] `Input` should have an `HTMLInputElement`
- [ ] `InnerInput` should have an `HTMLInputElement`
- [ ] `Password` should have an `HTMLInputElement`

Components using other components internally that have changed refs:
- [ ] `DatepickerRangeInput` using `Input`
- [ ] `PhoneInput` using `Input`

martimalek:
The draft PR ⬆️ solves the issue for `Input` and `Password`, `Textarea` is not forwarding a ref at the moment and `Datepicker` is forwarding one but it's not used AFAIK. [Our agreement on refs](https://github.com/freenowtech/wave/issues/156#issuecomment-880546292) talks about where to attach the ref to in a specific component but it does not say anything about in which components we should forward refs, should we do it in every form component? Should we do it only when necessary (when a user need arises)? In my opinion we could just forward refs for the components where they are being used `Input` and `Password` and drop it everywhere else (since imperative behaviour with refs should be avoided in general). Please give me your opinions... 🙏🏽

arturmiglio:
> and drop it everywhere else (since imperative behaviour with refs should be avoided in general). Please give me your opinions... 🙏🏽 

I'm not sure I agree with this.

Some doubts that maybe would also help others have a more informed opinion:
- What is the use case for ref in Input and Password (or any component for that matter)?
- Are there any the downsides of forwarding the ref?

Maybe some benchmarking would also help us decide? How other component libraries deal with the same topic?

martimalek:
I understand your concerns and agree that we lack some research to make an informed decision, I'll do the research and give an answer before the end of the week

martimalek:
Hey, I haven't had a lot of time to take a deep look into this, but I wanted to give an answer before the end of the week to at least give some initial context for everyone new to the discussion, so here it goes.

[...]

PR #328, opened 2023-04-26
fix: set focus on password input resolve #124

### LINKED ISSUE #124: Keep focus in the password field after the show/hide password button is clicked
At the moment, when the user clicks the eye button to show or hide the password, the input loses focus. It would be a better user experience if you can continue typing

raoasifraza11:
**What:**

Fixed set focus on password input while use click on eye button.

​
**Why:**

Improving the user experience, Closes #124 

​
**How:**

Added the focus based on input id.

​
**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/src/components/Password/Password.tsx
+++ b/src/components/Password/Password.tsx
@@ -96,6 +96,7 @@ const Password = forwardRef<HTMLDivElement, PasswordProps>(
                             onClick={() => {
                                 setIsHidden(prevValue => !prevValue);
 
+// TODO use ref passed to the input once https://github.com/freenowtech/wave/issues/169 is solved
                                 // set input focus
                                 const inputElement = document.querySelector(`input[id=${inputId}]`);
                                 if (inputElement) {
--- a/src/components/Password/Password.tsx
+++ b/src/components/Password/Password.tsx
@@ -98,7 +98,7 @@ const Password = forwardRef<HTMLDivElement, PasswordProps>(
 
 // TODO use ref passed to the input once https://github.com/freenowtech/wave/issues/169 is solved
                                 // set input focus
-                                const inputElement = document.querySelector(`input[id=${inputId}]`);
+                                const inputElement: HTMLElement = document.querySelector(`input[id=${inputId}]`);
                                 if (inputElement) {
                                     inputElement.focus();
                                 }
--- a/src/components/Password/Password.tsx
+++ b/src/components/Password/Password.tsx
@@ -95,6 +95,12 @@ const Password = forwardRef<HTMLDivElement, PasswordProps>(
                             type="button"
                             onClick={() => {
                                 setIsHidden(prevValue => !prevValue);
+
+                                // set input focus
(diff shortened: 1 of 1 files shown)


martimalek:
Hey @raoasifraza11, first of all thanks for using Wave and for taking your time contributing to it 🤗 

I've gone ahead and changed the title and description to reference #124 instead of #142, I guess it was a typo since #142 is already closed and does not relate to the PR as far as I see, please let me know in case you meant to reference #142

src/components/Password/Password.tsx
```diff
@@ -95,6 +95,12 @@ const Password = forwardRef<HTMLDivElement, PasswordProps>(
                             type="button"
                             onClick={() => {
                                 setIsHidden(prevValue => !prevValue);
+
+                                // set input focus
+                                const inputElement = document.querySelector(`input[id=${inputId}]`);
```
martimalek, line 100: I believe ideally we should use the `ref` that we are passing to the input instead of `querySelector`, but at the moment that won't work because the input's ref is actually being attached to the input wrapper instead of the actual input element (you can see it [in the `InnerInput` component](https://github.com/freenowtech/wave/blob/main/src/components/Input/InnerInput.tsx#L38)).

There is an issue to fix that (#169) but I'm not sure if we should wait for that to be resolved, since there may be users relying on the wrapper's ref and we may need to consider it a breaking change, I'd like to get other's opinions on this before making a decision

raoasifraza11:
> Hey @raoasifraza11, first of all thanks for using Wave and for taking your time contributing to it 🤗 
> 
> I've gone ahead and changed the title and description to reference #124 instead of #142, I guess it was a typo since #142 is already closed and does not relate to the PR as far as I see, please let me know in case you meant to reference #142

Sorry I meant #124

nlopin (approved):

div-Leo (approved):

src/components/Password/Password.tsx
```diff
@@ -95,6 +95,12 @@ const Password = forwardRef<HTMLDivElement, PasswordProps>(
                             type="button"
                             onClick={() => {
                                 setIsHidden(prevValue => !prevValue);
+
```
martimalek, line 98:
```suggestion

// TODO use ref passed to the input once https://github.com/freenowtech/wave/issues/169 is solved
```

martimalek (approved):

src/components/Password/Password.tsx
```diff
@@ -95,6 +95,13 @@ const Password = forwardRef<HTMLDivElement, PasswordProps>(
...
+// TODO use ref passed to the input once https://github.com/freenowtech/wave/issues/169 is solved
+                                // set input focus
+                                const inputElement = document.querySelector(`input[id=${inputId}]`);
```
martimalek, line 101:
```suggestion
                                const inputElement: HTMLElement = document.querySelector(`input[id=${inputId}]`);
```

PR #330, opened 2023-04-28
fix(tooltip): type children prop

### LINKED ISSUE #329: Tooltip typings do not allow children
-   `@freenow/wave` version: 1.30.0

### Relevant code

```ts
      <Tooltip content="Tooltip Content">
          Tooltip Trigger
      </Tooltip>
```

No TS errors. Instead, there is a following error:
`Type '{ children: Element; content: string; }' is not assignable to type 'IntrinsicAttributes & TooltipProps'.
  Property 'children' does not exist on type 'IntrinsicAttributes & TooltipProps'.ts(2322)`

### Reproduction

https://codesandbox.io/s/hardcore-scooby-yjxq49?file=/src/App.tsx

arturmiglio:
**What:**

Fix Tooltips props by typing children explicitly.

​
**Why:**

Close #329

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -3,12 +3,13 @@ import styled, { keyframes } from 'styled-components';
 import { usePopper } from 'react-popper';
 import { Placement } from '@popperjs/core/lib/enums';
 import { variant } from 'styled-system';
+
+import type { ReactNode } from 'react';
+
 import { MediaQueries, SemanticColors } from '../../essentials';
 import { get } from '../../utils/themeGet';
 import { Text } from '../Text/Text';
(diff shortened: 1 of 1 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -136,6 +138,10 @@ interface TooltipProps {
      * Force the tooltip to always be visible, regardless of user interaction
      */
     alwaysVisible?: boolean;
+    /**
+     * The node(s) passed as children will be used as the anchor point for the tooltip
+     */
+    children: ReactNode;
 }
 
 const Tooltip: React.FC<TooltipProps> = ({
```
nlopin, line 147: I think it is better to rely on React types and add children here instead of directly to the `TooltipProps`:
```suggestion
const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = ({
```

nlopin (approved):

Issue #335, opened 2023-05-22
Multiple links in InfoBanner


JanHamara:
### Describe the feature you'd like
Hello team, we currently have a need to render multiple links inside `<InfoBanner />` for the purposes of document cross-checking automation on DMT validation flow. Example below.
<img width="592" alt="Screenshot 2023-05-22 at 16 36 15" src="https://github.com/freenowtech/wave/assets/14894844/6eed937a-21a4-4b3b-aa81-62a8d9b8ec9c">

### Describe alternatives you've considered
Imo we have 3 options here:
 - Refactor `<InfoBanner>` component to accept an array of links to render below the banner description. `{ linkUrl: string, linkLabel: string }[]`
 - Refactor `<InfoBanner>` component to pass JSX with list of links setup via **children** prop
 - Create a snowflake in DMT for this purpose, if we don't want to integrate such feature into `<InfoBanner>` in our component library

div-Leo:
IMO the last two should not be valid options. 
If we keep adding snowflakes whenever we need something different from what we have in Wave we'll end up having a duplicated DS locally in our projects and most likely (not always) they won't reflect some possible new changes, in particular now with the rebranding.
The proposal in the second option is a pattern that we generally discourage to avoid again inconsistencies between projects. At the end of the day a DS is to delegate/enforce design decisions to it, if we let every component accept any kind of UI we'd break this intent.

The first option as it is will introduce a breaking change so I'd propose to accept `InfoLink | InfoLink[]` where `type InfoLink = { linkUrl: string, linkLabel: string }` and add a conditional check that either renders the link or the array of links.

WDYT?

JanHamara:
Yeah, totally agree with your point of view, just wanted to offer some alternatives. 👍 

Just not sure what you mean in last paragraph with the breaking change, where you would accept the `InfoLink | InfoLink []` ? There are two props at the moment, **`linkText`** and **`linkUrl`**, we should only have single prop, I suppose we agree on this, right?

nlopin:
I'm totally against the "pass array of links" approach. Today we need to allow a list of links. Tomorrow, we'll need to pass a subtitle. This will quickly fall into the mess of unrelated props we need to support.

We can keep the current level of granularity of the present component, and provide a more granular level API aside. We can do so by exposing a lower level components (e.g. `InfoBanner.Card`, `InfoBanner.Link`, etc) via a separate import.

[...]

PR #336, opened 2023-05-23
feat: implement compound component approach for InfoBanner

JanHamara:
**What:**
Implementing the compound components approach for building out the InfoBanner component (with multiple links as required for one of our products atm)
​
**Why:**
Explanation & Discussion: https://github.com/freenowtech/wave/issues/335

**How:**
- Create building blocks for InfoBanner: InfoBannerCard wrapper + blocks: Title, Description, LinkList (with defined spacing), Link
- Keeping the original component as is

**Media:**
Screenshot 2023-05-23 at 16 19 11

**Extra**
Added some temporary documentation, will document properly on Storybook after migration is complete, as this will go directly to **main** before migration.

**Checklist:**
-   [ ] Ready to be merged

### DIFF
--- a/src/components/InfoBanner/InfoBanner.tsx
+++ b/src/components/InfoBanner/InfoBanner.tsx
@@ -23,7 +23,7 @@ interface InfoBannerProps extends BoxProps {
     linkUrl?: string;
 }
 
-export type InfoBannerVariants = 'info' | 'success' | 'warning' | 'error';
+type InfoBannerVariants = 'info' | 'success' | 'warning' | 'error';
 
 interface BoxWithVariant extends BoxProps {
     variant: InfoBannerVariants;
--- a/src/components/InfoBanner/InfoBanner.tsx
+++ b/src/components/InfoBanner/InfoBanner.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import styled from 'styled-components';
 import { variant as styledVariant } from 'styled-system';
 import { get } from '../../utils/themeGet';
+import { theme } from '../../essentials/theme';
 import {
     CheckCircleSolidIcon,
     CloseCircleSolidIcon,
@@ -12,7 +13,8 @@ import {
 import { BoxProps, Box } from '../Box/Box';
 import { Link } from '../Link/Link';
 import { Text } from '../Text/Text';
-import { Colors, Spaces } from '../../essentials';
+import { Headline } from '../Headline/Headline';
+import { Spaces } from '../../essentials';
 
 interface InfoBannerProps extends BoxProps {
     title: string;
@@ -106,7 +108,7 @@ const emphasizedIconColorVariants = styledVariant({
     }
 });
 
-export const RoundedBox = styled(Box)<BoxWithVariant>`
+export const RoundedBox = styled(Box).attrs({ theme })<BoxWithVariant>`
     display: flex;
     flex-direction: row;
     justify-content: flex-start;
@@ -115,25 +117,18 @@ export const RoundedBox = styled(Box)<BoxWithVariant>`
     padding: ${`${Spaces[1]} ${Spaces[2]} ${Spaces[1]} ${Spaces[1]}`};
     ${({ emphasized }) => (emphasized ? emphasizedBannerVariants : bannerVariants)};
 
-    .Infobanner__Title,
-    .Infobanner__Description {
-        color: ${({ emphasized, variant }) =>
-            emphasized && variant !== 'warning' ? 'white' : Colors.AUTHENTIC_BLUE_900};
-    }
-
-    .Infobanner__Link {
-        &:link,
-        &:visited {
-            color: ${({ emphasized, variant }) =>
-                emphasized && variant !== 'warning' ? 'white' : Colors.ACTION_BLUE_900};
-        }
-
-        &:hover,
-        &:active {
-            color: ${({ emphasized, variant }) =>
-                emphasized && variant !== 'warning' ? Colors.AUTHENTIC_BLUE_350 : Colors.ACTION_BLUE_1000};
-        }
-    }
+    --info-banner-text-color: ${({ emphasized, variant }) =>
+        emphasized && variant !== 'warning'
+            ? get('semanticColors.text.primaryInverted')
+            : get('semanticColors.text.primary')};
+    --info-banner-link-color: ${({ emphasized, variant }) =>
+        emphasized && variant !== 'warning'
+            ? get('semanticColors.text.primaryInverted')
+            : get('semanticColors.text.link')};
+    --info-banner-link-hover-color: ${({ emphasized, variant }) =>
+        emphasized && variant !== 'warning'
+            ? get('semanticColors.text.tertiary')
+            : get('semanticColors.text.linkHover')};
 `;
 
 export const IconBox = styled(Box)<BoxWithVariant>`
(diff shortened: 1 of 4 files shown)


src/components/InfoBanner/blocks/InfoBannerCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import { ICON_VARIANTS, IconBox, InfoBannerVariants, ROLE_VARIANTS, RoundedBox } from '../InfoBanner';
+
+interface CardProps extends BoxProps {
+    children: ReactNode;
+    variant?: InfoBannerVariants;
+    emphasized?: boolean;
+}
+
+const LinkList = styled.ul`
```
jonotrujillo, line 16: Maybe this can be `List` instead of `LinkList`? Does not seem to require its children to be links.

src/components/InfoBanner/blocks/InfoBannerCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+    </Text>
+);
+
+const Description = ({ children }: { children: string }) => (
+    <Text className="Infobanner__Description" fontSize="small" textAlign="left">
+        {children}
+    </Text>
+);
+
+const Link = ({ linkUrl, linkText }: { linkUrl: string; linkText: string }) => (
```
jonotrujillo, line 38: To me it seem's like this link is coupled to a very specific behaviour. For example, I'm not able to pass an `onClick` prop. But maybe that's fine given it's a compound component. What do you think?

src/components/InfoBanner/blocks/InfoBannerCard.tsx
JanHamara: The styling of this LinkList (padding) is actually quite coupled to the specs we are trying to meet (with links) with the addition of these compound components, as mentioned below in other response, I am not sure we would want the same for e.g. textual content in list items, that would naturally go into `InfoBannerCard.Description`.

Personally I would prefer to keep this naming to make that clear, that we don't want just anything to be put in there :)

martimalek, line 16: I'm wondering if `LinkList` should even exist... I believe the requirement to have a list of links is only for DMT and to my understanding we should only be exposing the components that make the `InfoBanner` (meaning Card, Title, Description and Link).

We'd still be allowing to have a list of links (or whatever the user needs) with the `Link` styled according to our `InfoBanner` specs but the `LinkList` would have to be implemented in DMT, this way we still offer flexibility without coupling Wave with DMT.

Otherwise if a different project needs a list of descriptions wouldn't we have to create a `DescriptionList` even though the `InfoBanner` doesn't really use it?

nlopin: I agree with Martí, I think this should be a custom DMT thing

src/components/InfoBanner/InfoBanner.tsx
```diff
@@ -106,21 +106,41 @@ const emphasizedIconColorVariants = styledVariant({
...
+export const RoundedBox = styled(Box)<BoxWithVariant>`
     display: flex;
     flex-direction: row;
     justify-content: flex-start;
     border-radius: ${get('radii.3')};
     border: 0.0625rem solid;
     padding: ${`${Spaces[1]} ${Spaces[2]} ${Spaces[1]} ${Spaces[1]}`};
     ${({ emphasized }) => (emphasized ? emphasizedBannerVariants : bannerVariants)};
+
+    .Infobanner__Title,
```
nlopin, line 118: I propose to use the inversion of control here. Instead of cascading, we set CSS variables for colours. Building blocks use those colours, but they set their style:
image

src/components/InfoBanner/blocks/InfoBannerCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+    margin-block-start: 0.3rem;
+    margin-block-end: 0.3rem;
+    padding-inline-start: 1.5rem;
+
+    li {
+        line-height: 1.1;
+    }
+`;
+
+const Title = ({ children }: { children: string }) => (
```
nlopin, line 26: It looks like we always had an issue that the title `span`, but ideally, it should be flexible and maybe a headline by default. Created a bug #338

src/components/index.ts
```diff
@@ -55,3 +55,4 @@ export { SelectListProps } from './SelectList/types';
 export { Row, RowProps, Column, ColumnProps } from './Grid/Grid';
 export { Popover, PopoverProps } from './Popover/Popover';
 export { InfoBanner, InfoBannerProps, InfoBannerVariants } from './InfoBanner/InfoBanner';
+export { InfoBannerCard } from './InfoBanner/blocks/InfoBannerCard';
```
nlopin, line 58: How does the import look when using the library?

nlopin (approved):
If you are in a rush, I think it is fine to merge and then refine the implementation

src/components/InfoBanner/blocks/InfoBannerCard.tsx
JanHamara: ok, I removed it 👍

JanHamara: has been removed now 👍

src/components/InfoBanner/InfoBanner.tsx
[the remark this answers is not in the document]
JanHamara: Do you mean to completely remove original API and thus original props, or just using new compound components to render title, desc, etc. in the main component from original props? 

I think if we keep original props and just utilise new compound components in main comp., it's still sort of two different implementations, cuz I would ideally want to remove the original props. But if we do so, then we would introduce breaking changes, having to update all current utilisations of InfoBanner in our products, which I don't really want to do with this MR.

src/components/InfoBanner/blocks/InfoBannerCard.tsx
JanHamara: I will close this now to move the MR, but if we feel need to add an `onClick` on `<Link>` component, we can create an issue for it then and update it here :)

src/components/InfoBanner/blocks/InfoBannerCard.tsx
JanHamara: I suppose we can change it to `<Headline>` in this MR

src/components/index.ts
JanHamara: being exported here, it should be from `import { InfoBannerCard } from '@freenow/wave'` right?

src/components/InfoBanner/InfoBanner.tsx
JanHamara: Nice suggestion, I've done that 👍

src/components/index.ts
JanHamara: As said below, happy to refine the implementation and export strategy later, I implemented most of the changes from MR suggestions apart from touching the original API and export here.

src/components/InfoBanner/InfoBanner.tsx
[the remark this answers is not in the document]
JanHamara: Also as mentioned below, bit urgent MR, so I will plan to revisit this and refine the original component, for now I will proceed with the needed separate compound components.

martimalek (approved):

src/components/InfoBanner/blocks/InfoBannerCard.tsx
div-Leo: I'm definitely late to the conversation but that might be a problem if we wanted to add some tracking for example.
We could have exposed the `StyledLink` itself instead of wrapping it, because intrinsically it accepts a `children` a
`href` and also all the props from `WaveLink` 

```tsx
InfoBannerCard.Link = StyledLink;
```

Sorry for being late 😅

PR #340, opened 2023-06-09
fix: add support for children props inside Label component

rasidre:
**What:**

Add support for children props inside Label component.

​
**Why:**

When using the Label component, the console shows following error:
```Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes & LabelProps'.```

​
**How:**

Extend the LabelProps with ```ComponentPropsWithoutRef<'span'>```

### DIFF
--- a/src/components/Label/Label.tsx
+++ b/src/components/Label/Label.tsx
@@ -1,11 +1,12 @@
 import styled from 'styled-components';
 import { compose, margin, MarginProps, ResponsiveValue, variant } from 'styled-system';
+import { ComponentPropsWithoutRef } from 'react';
 import { Text } from '../Text/Text';
 import { Colors } from '../../essentials';
 import { theme } from '../../essentials/theme';
 import { get } from '../../utils/themeGet';
 
-interface LabelProps extends MarginProps {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/components/Label/Label.tsx [hunk 1]
```diff
@@ -1,11 +1,12 @@
 import styled from 'styled-components';
 import { compose, margin, MarginProps, ResponsiveValue, variant } from 'styled-system';
+import { ComponentPropsWithoutRef } from 'react';
 import { Text } from '../Text/Text';
 import { Colors } from '../../essentials';
 import { theme } from '../../essentials/theme';
 import { get } from '../../utils/themeGet';
 
-interface LabelProps extends MarginProps {
+interface LabelProps extends ComponentPropsWithoutRef<'span'>, MarginProps {
```
nlopin, line 9: shouldn't it be `'span'`?

rasidre: @nlopin: That's what the associated test attempts to do. 
```
it('renders an <span> tag by default', () => {
    expect(render(<Label />)).toMatchHtmlTag('span');
});
```

nlopin: Got it 👍

nlopin (approved):

PR #347, opened 2023-07-26
fix: support children prop type in row component

### LINKED ISSUE #345: Row component missing children prop type definition
-   `@freenow/wave` version: 1.32.0

### Relevant code

```js
<Row>Some content</Row>
```

The following TypeScript error **not** to be shown:

```
Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes & BoxProps'
```

I think it's an issue similar to #329 and #332.

### Reproduction

https://codesandbox.io/s/wave-playground-forked-nl8lr5?file=/src/App.tsx

jonotrujillo:
**What:**

Add missing `children` prop type definition to `Row` component.

Fixes #345.

**Why:**

`children` prop type was not supported for `Row` component.
​
**How:**

Extending component props with `PropsWithChildren`.
​
**Media:**

N/A

**Checklist:**

-   [x] Ready to be merged

### DIFF
--- a/src/components/Grid/Grid.tsx
+++ b/src/components/Grid/Grid.tsx
@@ -16,7 +16,7 @@ const BorderBoxWrapper = styled(Box)`
 
 type RowProps = BoxProps;
 
-const Row: FC<RowProps> = (props: RowProps) => (
+const Row: FC<PropsWithChildren<RowProps>> = props => (
     <Box display="flex" flexWrap="wrap" marginRight={`-${GAP}`} {...props} />
 );
 


martimalek (approved):
LGTM.

Maybe outside the scope of the PR, do we have any opinions on using `FC`? Should we get rid of them and start typing only props?

nlopin (approved):

nlopin:
> LGTM.
> 
> Maybe outside the scope of the PR, do we have any opinions on using `FC`? Should we get rid of them and start typing only props?

We need to choose one approach, currently there are components that do both `const Component: FC<Props> = ({...rest}: Props) {...}`

martimalek:
IMO we should only type the props, the component will implicitly be typed as `JSX.Element`, by getting rid of `FC` we will also make the migration to React 18 simpler, since `FC` has implicit children in 17 but not in 18. Check [other downsides of `FC`](https://github.com/facebook/create-react-app/pull/8177).

PR #350, opened 2023-07-31
Render Tooltip via React Portal

### LINKED ISSUE #349: Tooltip causes scrollbar in Modal dialog
We've updated wave from 1.15.1 to 1.32.1 (latest) and we are facing an issue with tooltips causing a scrollbar to appear when they are being rendered within a Modal dialog.

image

-   `@freenow/wave` version: 1.32.1

### Relevant code

```js
<Modal>
  <Tooltip
    content="Tooltip text"
    placement="center-right"
  >
    ...
  </Tooltip>
</Modal>
```

image
The tooltip was expected to overflow the modal.

The user should see the whole content of the tooltip, they should not be scrolling to see it.

[internal discussion](https://mytaxi.slack.com/archives/CLM50RECU/p1690799683872489)

jahe:
**What:**

In this PR we're adding a new prop `container` to the `Tooltip` component where the tooltip should be rendered into.
​
**Why:**

This PR closes #349.

Tooltips that are rendered in Modal dialogs can cause scrollbars to appear. It was not like that in previous version of wave (1.15.1).

**How:**

* Introduce new (optional) prop `container` to Tooltip component where the tooltip body should be rendered into
* Render the tooltip body into `container` via react portal
* See [internal discussion](https://mytaxi.slack.com/archives/CLM50RECU/p1690799683872489) 
* Implementation according to [popper library documentation](https://popper.js.org/react-popper/v2/react-portals/) 
​
**Media:**

Before:
image

After:
image

**Checklist:**

-   [ ] Ready to be merged

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -7,7 +7,7 @@ import { variant } from 'styled-system';
 
 import type { PropsWithChildren } from 'react';
 
-import { MediaQueries, SemanticColors } from '../../essentials';
+import { Elevation, MediaQueries, SemanticColors } from '../../essentials';
 import { get } from '../../utils/themeGet';
 import { Text } from '../Text/Text';
 import { mapPlacementWithDeprecationWarning, TooltipPlacement } from './TooltipPlacement';
@@ -90,6 +90,7 @@ interface TooltipBodyProps {
 
 const TooltipBody = styled.div<TooltipBodyProps>`
     position: relative;
+    z-index: ${Elevation.TOOLTIP};
     background-color: ${p =>
         p.inverted ? SemanticColors.background.secondary : SemanticColors.background.primaryEmphasized};
     padding: 0.25rem 0.5rem;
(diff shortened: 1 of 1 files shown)


nlopin:
What do you think if we create a portal by default and attach it to the `document.body`? 
Our users won't even face the issue anymore in that case, the whole issue disappear. I don't see any negative sides on this solution

nlopin (changes requested):
What do you think if we create a portal by default and attach it to the `document.body`? 
Our users won't even face the issue anymore in that case, the whole issue disappear. I don't see any negative sides on this solution

jahe:
Good idea. We would need to put a big enough z-index on the tooltip then so that it gets rendered on top of modals. With my changes the responsibility for that is delegated to the user of the library which is not very convenient. I agree that it would be better if the library makes sure that it "just works".

> create a portal by default
Do you mean using `document.body` as portal to render tooltips in any case instead of doing it conditionally? In that case we could remove the `container` prop.

nlopin:
> Do you mean using `document.body` as portal to render tooltips in any case instead of doing it conditionally? In that case we could remove the `container` prop.

Yes, I'd do that. In that case the z-index won't gonna be necessary because the tooltip will be sitting on top of the tree and in a separate stacking context

jahe:
The issue is that Modals come with `z-index: 1100` which makes them appear on top of elements that are rendered in a separate stacking context ([see example](https://codesandbox.io/s/mutable-water-3ltfr5?file=/index.html)).

jahe:
The issue is that Modals come with `z-index: 1100` which makes them appear on top of elements that are rendered in a separate stacking context ([see example](https://codesandbox.io/s/mutable-water-3ltfr5?file=/index.html)).

jahe:
I've removed the `container` prop in 94b053c. Tooltips are rendered via React Portal into `document.body` now.

nlopin (approved):

div-Leo (approved):

Issue #324, opened 2023-04-19
Actionable Breadcrumbs 


div-Leo:
### Describe the feature you'd like

Add a new Breadcrumbs component as in the image.
Each step should be a link that redirects you to the relative page.
<img width="335" alt="Screenshot 2023-04-19 at 09 15 33" src="https://user-images.githubusercontent.com/12762609/232996856-8c456ebc-1db3-4bb8-8b07-de7197cebc9e.png">

Considerations:
We need to understand how to integrate it with different router providers or just provide the building block and delegate the navigation to the user.

### Describe alternatives you've considered

We could add a snowflake inside the project where we need this feature, but it might also be useful for other projects.

arturmiglio:
Resolved by #353

PR #353, opened 2023-08-11
feat: add breadcrumbs component

div-Leo:
**What:**

​
**Why:**

Add new component Breadcrumbs
​
**How:**

Created new component
Added documentation
Wrote tests

**Media:**
Capture-2023-08-11-144009

Issue:  #324

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -46,11 +46,13 @@ const Breadcrumbs = ({ children, inverted }: BreadcrumbsProps): JSX.Element => {
         <BreadcrumbsList>
             {Children.map(arrayChildren, (child, index) => (
                 <BreadcrumbsListItem>
-                    {cloneElement(child as ReactElement, {
-                        inverted
-                    })}
+                    <nav aria-label="breadcrumbs">
+                        {cloneElement(child as ReactElement, {
+                            inverted
+                        })}
+                    </nav>
                     {index < arrayChildren.length - 1 ? (
-                        <Box mx="0.25rem" height={16} mt="0.125rem">
+                        <Box height={16} mt="0.125rem">
                             <ChevronRightIcon size={16} color={Colors.AUTHENTIC_BLUE_350} />
                         </Box>
                     ) : // eslint-disable-next-line unicorn/no-null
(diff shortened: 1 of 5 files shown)


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,90 @@
...
+`;
+
+const BreadcrumbsListItem = styled.li`
+    display: flex;
+`;
+
+const Breadcrumbs = ({ children, inverted }: BreadcrumbsProps): JSX.Element => {
+    const arrayChildren = Children.toArray(children);
+
+    return (
```
nlopin, line 45: It is recommended to have the breadcrumb wrapped with the `nav` tag for a11y. 
I found a very good example on the [a11y style guide website](https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-breadcrumbs) we can use as a markup example.

- [ ] follow the a11y recommendations

src/components/Breadcrumbs/docs/Breadcrumbs.storybook.mdx
nlopin, line 1: How do we use `Breadcrumbs` with the react-router or next links? Let's show them examples

nlopin (changes requested):
We need to add accessibility to the component.

Another question: can we make the clickable area bigger? At the moment only the text is clickable, why not having the whole block clickable? I think it can be achieved by wrapping the `li` with `a`.

Now:
CleanShot 2023-08-14 at 13 38 03@2x

Ideally:
CleanShot 2023-08-14 at 13 39 16@2x

src/components/Breadcrumbs/docs/Breadcrumbs.storybook.mdx
div-Leo: We already have an example of how to use it with react-router, is that enough or do we want to add also Next Links?

div-Leo:
@nlopin I couldn't wrap the `li` with the `a` because of the structure I chose, it was easier to use padding to increase the area, and the result should be okay.

Screenshot 2023-08-17 at 10 07 21

src/components/Breadcrumbs/docs/Breadcrumbs.storybook.mdx
nlopin: I missed that, sorry

nlopin (approved):

Issue #357, opened 2023-08-24
Codemod to migrate from hardcoded colors to css variables


nlopin:

martimalek:
Tackled in #363

PR #363, opened 2023-08-25
Create codemod to migrate colors to css variables

martimalek:
**What:**

Adds a codemod that migrates usages of the old `Colors` constant to CSS variables following our new colors structure

​
**Why:**

To ease the adoption of Wave 2.0 by automating the colors migration

**How:**

- Add tests for both the existing codemod and the new one
  - In the __testfixtures__ folders there are `x.input.tsx` with their respective `x.output.tsx` (the output file is what the input file should look like after running the codemod on it)
  - In the __tests_ I simply define the path and the config to run the codemods tests
- Rename existing `v2` codemod to more semantic `block-to-width-100`
- Add new `colors-to-css-vars` codemod

**Checklist:**

-   [x] Ready to be merged

I'll open another PR updating the docs related to codemods and moving the /codemods folder to add it to the build of the project

Relates to #357


nlopin (approved):
Do you think we can put a comment to the code recommending switching to semantic colours via `getSemanticValue()`?

martimalek:
> Do you think we can put a comment to the code recommending switching to semantic colours via `getSemanticValue()`?

I think that can be tricky, when using the [AST explorer](https://astexplorer.net/) it seems that comments don't always represent nodes in the AST, and when they do, they don't have a value 😕 (see attached). Looking at jscodeshifts API it seems that some nodes have a `comments` or `leadingComment` property, so I think it is possible to do, but if it's okay with you I'd prefer to have a comment in the migration guide with the recommendation for now and focus on other issues of the 2.0 milestone, in case we are ready for release and still have more time this can be a nice to have. 
image

Issue #366, opened 2023-08-29
Fix dependency conflicts and TS issues in next branch


martimalek:
We have multiple dependency conflicts and TS issues in the `next` branch, we must fix them before we can release our next major.

TS issues
<img width="709" alt="image" src="https://github.com/freenowtech/wave/assets/46452321/3ba873e6-cd06-4175-9e05-9face73009f4">

One of the dependency conflicts for reference
<img width="690" alt="image" src="https://github.com/freenowtech/wave/assets/46452321/0c96ad41-af63-4c0b-a2dc-43e95c2670cd">

PR #367, opened 2023-08-29
Fix ts and dependency issues

martimalek:
**What:**

Fixes TS issues that made the build fail and dependency conflicts that forced the install with the `--legacy-peer-deps` flag (except a conflict with `react-tether`, which will be solved separately).

​
**Why:**

To have a working build and cleaner deps.

​
**How:**

- To fix the build
  - Import missing `theme` in `InfoBanner.tsx`
  - Add necessary casts in `BoxedInput.tsx`
- To (almost) fix the deps conflicts
  - Migrate from `@testing-library/react-hooks` to `@testing-library/react` (the same APIs are now exposed from `@testing-library/react` so the former is not needed anymore)
  - Remove unused `theme-ui` package
  - Bump `ts-jest` from `^25.2.1` to `^26.5.6`
  - Bump `react-windowed-select` from `^4.0.0` to `^5.1.0`

**Checklist:**

- [x] Check `react-tether` TS issue in pipeline
- [x] Ready to be merged

After this PR the build issues are fixed and the only remaining dependency conflict is with `react-tether`, which will be removed from the project in favour of `react-popper` in #232 

Relates to #232
Relates to #366 (will be closed together with #232)


martimalek:
I tried bumping `react-tether` to 3.x.x but there are some TS issues that didn't seem easy to solve, IMO we should directly tackle #232

src/components/Input/BoxedInput.ts
```diff
@@ -81,7 +83,11 @@ const BoxedInput: FC<InternalInputComponentProps> = styled(BaseInput)<InternalIn
                 `linear-gradient(0deg, ${getSemanticValue(
                     // inverted doesn't work well for the modern layout
                     p.inverted ? 'background-element-primary-default' : 'background-element-neutral-default'
-                )} calc(50% + ${p.size === 'small' ? '0.0825rem' : '0.0625rem'}), transparent 50%)`};
+                )} calc(50% + ${
+                    (p.size as Pick<InternalInputComponentProps, 'size'>['size']) === 'small'
```
jonotrujillo, line 87: nitpick: can also be written as `p.size as InternalInputComponentProps['size']`

jonotrujillo (approved):

PR #368, opened 2023-08-29
feat: add inverted prop codemod

### LINKED ISSUE #365: Codemod to migrate from `inverted` prop to `InvertedColorScheme ` component
## The context

With the introduction of the dark theme in Wave (#356), we decided to remove `inverted` props from the components where they existed:
* `Input`
* `Password`
* `Textarea`
* `Button`
* `Select`
* `SelectList`
* `PhoneInput`
* `Datepicker`
* `Tooltip`
* `Breadcrumbs`
* `Logo`

The inverted colours were used to assemble the dark theme, which means the dark theme is fully compatible with the previous look and feel.

## Problem

We have users of the `inverted` prop who must migrate their code to the new version. They must delete the `inverted` prop and wrap the piece of the interface with the `FlipTheme` component to switch it to the dark mode:

**Before**

```jsx
<Box as="form">
  <Input name="hello" inverted />
  <Textarea inverted />
  <Button type="submit" inverted>Submit</Button>
</Box>
```

**After:**

```jsx
<FlipTheme>
  <Box as="form">
    <Input name="hello" />
    <Textarea />
    <Button type="submit">Submit</Button>
  </Box>
</FlipTheme>
```

## Solution

In the ideal case, our code mode can detect components with the `inverted` prop, find their common parent and wrap it with the `FlipTheme` component. At the same time, it requires precision and knowledge about the product UI, so I doubt we can provide a good migration experience and not break the UI at the same time by putting the wrapper too high up the component tree.

**The solution** with the okay-ish migration experience is to _work on the component level and wrap each component with the `inverted` prop with the `FilpTheme` component._ The downside of this approach is the potential decline in performance (the DOM becomes deeper). 

We should highlight this issue in the migration guide and recommend refactoring the migrated code to keep as few `FlipTheme` components as possible

martimalek:
**What:**

Adds a codemod that replaces the `inverted` prop for the `FlipTheme` component
​
**Why:**

To ease the adoption of Wave 2.0
​
**How:**

- Add new `inverted-to-flip-theme` codemod
- Add tests for the new codemod

**Checklist:**

- [ ] The FlipTheme component does not yet exist
- [ ] Ready to be merged

Closes #365

### DIFF
--- a/docs/migrating.storybook.mdx
+++ b/docs/migrating.storybook.mdx
@@ -50,3 +50,15 @@ This won't change how your icons look in any way since we already exported the d
 ```bash
 npx jscodeshift -t node_modules/@freenow/wave/lib/cjs/codemods/deprecated-icons.js path/to/src
 ```
+
+### Inverted property deprecation
+
+With the introduction of our new dark theme in Wave we've removed the `inverted` property that several components had. In case you want a specific component to have
+the styling of the opposite theme you need to wrap it with the `FlipTheme` component.
+
+```bash
+npx jscodeshift -t node_modules/@freenow/wave/lib/cjs/codemods/inverted-to-flip-theme.js path/to/src
+```
+
+Disclaimer: This codemod wraps every Wave component that is using the `inverted` property with `FlipTheme`, this could lead to a decline in performance in case the
+DOM gets too deep. Ideally when you have Wave components using the `inverted` prop with a common parent you wrap the parent instead of each individual component.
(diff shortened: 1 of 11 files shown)


src/codemods/__tests__/block-to-width-100-test.ts
```diff
@@ -0,0 +1,6 @@
+// import { defineTest } from 'jscodeshift/dist/testUtils'
```
jonotrujillo, line 1: I think this can be removed?

jonotrujillo (approved):

martimalek:
I've started improving the codemod to identify siblings that use the `inverted` prop to wrap them together for better performance. The work is in branch [inverted-codemod-improvements](https://github.com/freenowtech/wave/tree/inverted-codemod-improvements) in commit [b9db2dc](https://github.com/freenowtech/wave/commit/b9db2dcddc5e8f775ffff643b3157b7577acb6a8). I've decided to park it for a bit because it's more complex than what I initially thought, I'll go back to it when all the remaining codemods for Wave 2.0 are done.

nlopin (approved):

Issue #354, opened 2023-08-17
Error Variant for Tag and SelectList Options


jahe:
Hello 👋

As part of [RP-10976](https://jira.intapps.it/browse/RP-10976) we would like to add a new variant for "error" state to the [Tag](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?type=design&node-id=237-7318&mode=design&t=vBRlutCS6iPjIQnL-0) and the [SelectList](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?type=design&node-id=310-0&mode=design&t=vBRlutCS6iPjIQnL-0) components in wave. The design in figma is [ready](https://www.figma.com/file/dTefUItBx5bfbxcCcMzMXm/Wave-DS?type=design&node-id=237-7318&mode=design&t=vBRlutCS6iPjIQnL-0).

**Proposed Changes**
* Add new prop "variant" to Tag component ([see code](https://github.com/freenowtech/wave/blob/c67d40e939b4a79c91b7a0cef4fe314a2eac59e6/src/components/Tag/Tag.tsx#L11-L11), [see example](https://github.com/freenowtech/wave/blob/c67d40e939b4a79c91b7a0cef4fe314a2eac59e6/lib/types/components/Banner/Banner.d.ts#L7-L7)) and apply the styles accordingly ([see code](https://github.com/freenowtech/wave/blob/c67d40e939b4a79c91b7a0cef4fe314a2eac59e6/src/components/Tag/Tag.tsx#L42-L42))
* Apply custom styles for SelectList option by adjusting StylesConfig ([see code](https://github.com/freenowtech/wave/blob/c67d40e939b4a79c91b7a0cef4fe314a2eac59e6/src/components/SelectList/SelectList.tsx#L137-L137))

jahe:
**Examples**

![image](https://github.com/freenowtech/wave/assets/2736136/495b507d-8c37-4886-b3a1-15fecc1bca5b)
![image](https://github.com/freenowtech/wave/assets/2736136/5ee8f4d4-9eb8-473a-b709-1ec1d7220d7d)

nlopin:
@jahe could you clarify this task a bit? Without the context, It looks like we shouldn't allow the user to select an error option

PR #372, opened 2023-08-30
Add variants of tag component in Tag and SelectList

hackbnw:
**What:**

 Error Variant for Tag and SelectList Options (#354).
​
**Why:**

We want to highlight certain items to make it clear to the users that those are invalid and they should remove them. Otherwise the configuration will be considered broken.
​
**How:**

- Fix missing Open Sans font in storybook
- Add storybook stories for SelectList with multi select
- Support new `variant = "default" | "disabled" | "error"` property in the Tag component
- Support having "error" property on options of SelectList element to render the entry tag in "error" variant
- Use semantic colors in accordance with design specification for tags in Tag and SelectList
​

**Media:**

Tag:

default

disabled

error

SelectList:

image

image

**Checklist:**

-   [ ] Ready to be merged

### DIFF
--- a/src/components/SelectList/SelectList.tsx
+++ b/src/components/SelectList/SelectList.tsx
@@ -21,6 +21,19 @@ import { SelectListProps } from './types';
 
 type WithSelectProps<T> = T & { selectProps: SelectListProps };
 
+const getOptionError = (option: unknown): boolean =>
+    typeof option === 'object' && 'error' in option && Boolean(option.error);
+
+const getOptionVariant = (selectProps: Props, option: unknown): 'default' | 'disabled' | 'error' => {
+    if (selectProps.isDisabled) {
+        return 'disabled';
+    }
+
+    return !getOptionError(option) ? 'default' : 'error';
+};
+
+const getColor = (key: string, props: Props) => String(get(key)(props));
+
 const customStyles: StylesConfig = {
     container: (provided, { selectProps }: WithSelectProps<Props>) => {
         const bSize = {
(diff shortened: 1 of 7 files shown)


jahe (approved):
🎉

src/components/SelectList/SelectList.tsx
```diff
@@ -21,6 +21,19 @@ import { SelectListProps } from './types';
...
 
+const getOptionError = (option: unknown): boolean =>
+    typeof option === 'object' && 'error' in option && Boolean(option.error);
+
+const getOptionVariant = (selectProps: Props, option: unknown): 'default' | 'disabled' | 'error' => {
+    if (selectProps.isDisabled) {
+        return 'disabled';
+    }
+
+    return !getOptionError(option) ? 'default' : 'error';
```
nlopin, line 32: We should swap it to keep the "else" branch always showing the default variant. It will be easier to have the default variant chosen in all the non-specific cases

src/components/SelectList/SelectList.tsx
```diff
@@ -169,35 +182,79 @@ const customStyles: StylesConfig = {
...
     },
-    multiValue: (provided, { selectProps }: { selectProps: Props }) => {
+    multiValue: (provided, { selectProps, data }) => {
+        const optionVariant = getOptionVariant(selectProps, data);
+
         const styles = {
             ...provided,
-            color: Colors.ACTION_BLUE_900,
-            border: `0.0625rem solid ${Colors.ACTION_BLUE_900}`,
+            border: `0.0625rem solid`,
```
nlopin, line 190:
```suggestion
            border: '0.0625rem solid',
```

nlopin (changes requested):

nlopin (approved):

PR #374, opened 2023-09-19
feat: add codemod for weak prop replacement in text

### LINKED ISSUE #369: Create codemod to replace `weak` prop in Text
The `Text` component has a `weak` prop which is deprecated in favour of `secondary`.

We should create a codemod to ease the migration and remove the `weak` prop in Wave 2.0.

martimalek:
**What:**

Adds a codemod that replaces the deprecated `weak` prop in the `Text` component for the `secondary` prop

​
**Why:**

The `weak` prop is deprecated and we want to ease the adoption of Wave 2.0

​
**How:**

- Add new `weak-to-secondary` codemod
- Cover with tests
​

**Checklist:**

- [x] Remove `weak` prop from `Text`
- [x] Add new codemod in migrating docs
- [x] Ready to be merged

Closes #369


martimalek:
> Do we handle the case when the prop is set dynamically?
> 
> ```
>     <WaveText weak={Date.now() % 2 === 0}>
>         Just a text
>     </WaveText>
> ```

Nice catch!! I hadn't thought of the possibility, I've just pushed a new commit covering the case and simplifying a bit the codemod

PR #375, opened 2023-09-19
refactor: replace deprecated weak prop for secondary in text usages

### LINKED ISSUE #304: Remove 'weak' prop usages in Wave components
-   `@freenow/wave` version: 1.29.0

### What is the problem?

We are still using the `weak` prop from Text component (which has been deprecated in favour of `secondary` prop) in some of our own Wave components in the repo, which causes the console error to throw on our products, even if the product itself has refactored all `weak` prop occurrences to `secondary`.

Screenshot 2023-02-28 at 17 31 35

### Relevant Components

- FilePicker

Also...
- ModalCreator (in Modal/docs)
- OffcanvasCreator (in Offcanvas/docs)
- Pagination/docs .mdx file
- PaginationProvider in /docs
- ComplexDataTable example (in Table/docs)
- LiveTable example (in Table/docs)
- SkeletonTable example (in Table/docs)
- SpacesContainer (in Spaces/docs)
- IconGrid (in icons/docs)

And then of course in Text component and TextPropsTable (in Text/docs) but that can only be removed after all products have changes their `weak` prop usage with `secondary`

### Solution

The main action is to remove the usage of `weak` prop from **FilePicker** component, but we should remove all occurences in other relevant files if possible now.

martimalek:
**What:**

Replace the deprecated `weak` prop in favour of `secondary` in internal usages of the `Text` component

​
**Why:**

The `weak` prop is deprecated so we shouldn't use it, on top of that using it throws a warning which bubbles up to external projects that use Wave components with usages of `weak`

​
**How:**

Replace all usages of `weak` prop of `Text` components

**Checklist:**

-   [x] Ready to be merged

Closes #304

### DIFF
--- a/src/components/FilePicker/FilePicker.tsx
+++ b/src/components/FilePicker/FilePicker.tsx
@@ -100,17 +100,17 @@ const Outliner = styled(Box)<OutlinerProps>`
     ${({ hasValidFile }) =>
         hasValidFile &&
         css`
-                ${MediaQueries.medium} {
-                    &:hover {
-                        background-color: ${Colors.ACTION_BLUE_50};
-                        border-color: ${Colors.ACTION_BLUE_50};
+            ${MediaQueries.medium} {
+                &:hover {
+                    background-color: ${Colors.ACTION_BLUE_50};
+                    border-color: ${Colors.ACTION_BLUE_50};
 
-                        svg:not([color='${Colors.POSITIVE_GREEN_900}']) path {
-                            fill: ${Colors.ACTION_BLUE_900};
-                        }
+                    svg:not([color='${Colors.POSITIVE_GREEN_900}']) path {
+                        fill: ${Colors.ACTION_BLUE_900};
                     }
                 }
-            `}
+            }
+        `}
 `;
 
 const Input = styled.input`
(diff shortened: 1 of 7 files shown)


nlopin:
Let's remove the `weak` from the component code as well

martimalek:
> Let's remove `weak` from the component code as well

Hey @nlopin, I'm not sure if I follow, I was planning on removing the `weak` prop altogether in a different PR branching out from `next`, since the current PR is pointing to `main` it will be released in the current major and removing `weak` would be a breaking change, did I misunderstand you?

I could use the same PR for the codemod (#374) to remove the `weak` prop if you prefer...

nlopin:
> > Let's remove `weak` from the component code as well
> 
> Hey @nlopin, I'm not sure if I follow, I was planning on removing the `weak` prop altogether in a different PR branching out from `next`, since the current PR is pointing to `main` it will be released in the current major and removing `weak` would be a breaking change, did I misunderstand you?
> 
> What I could do is use the same PR for the codemod (#374) to remove the `weak` prop for instance...

My bad! it should be removed in the `next` branch, not `main`. Codemod PR will do

PR #379, opened 2023-09-26
feat: adding aria-invalid when an input has an error for accessibility

lloydaf:
**What:**

In this PR we are adding `aria-invalid` to the `Input` component if there is an error
​
**Why:**

​This is for better accessibility, currently we can only assert if an input box contains an error through CSS computations, which is not the best practice
**How:**

Add `aria-invalid` if `error` exists
​
**Media:**

**Checklist:**

-   [x] Ready to be merged


nlopin (approved):
Great work!

PR #380, opened 2023-09-27
resolve conflicts between next and main

martimalek:
**What:**

Resolve conflicts between `next` and `main`
​
**Why:**

We want to have `next` up to date with `main`
​
**How:**

- Create new branch from `next` and merged `main` into it
- Accepted all incoming changes except in the `SelectList` and `Tag` components, where the migration to css variables had already been done, so I accepted kept that (new variants for those components lack css vars, I'll open a PR soon)

-   [x] Ready to be merged

> **Be aware!** There are build and test issues, but those will get resolved when taking care of migrating to css variables the `Tag`, `SelectList` and `Breadcrumbs` components (issues #373 and #361). I'll open a PR for that before EOD.
Alternatively, I can do those changes in this MR to keep `next` clean, let me know if you prefer that.

Thanks to having next up to date I'll be able to update #374, removing the `weak` prop altogether and merging the PR.

PR #381, opened 2023-09-27
Feat/migrate tag and selectlist css vars

### LINKED ISSUE #373: Migrate Tags and SelectList component updates to css variables
After #372 is merged, we need to update `Tag` and `SelectList` to match the new theme and css variables

martimalek:
**What:**

Migrate new variants of `Tag` and `SelectList` components to use css vars for colors.

​
**Why:**

The `Tag` and `SelectList` components have recently received some updates in `main`, since in `main` the css variables are not yet available these updates were done using the "old" colors constants. We need to migrate to the new css vars here in `next`.

​
**How:**

I've migrated all color values in `Tag` and `SelectList` so they are using `getSemanticValue`. Please review the semantic values that I've chosen and let me know if you think that there are better alternatives, as far as I understand, we don't have a way to perfectly map the old semantic colors to the new ones, I've tried to keep them semantic.

​
**Media:**

`Tag`

image

`SelectList` (multi)

image

**Checklist:**

-   [x] Ready to be merged

Closes #373

### DIFF
--- a/src/components/SelectList/SelectList.spec.tsx
+++ b/src/components/SelectList/SelectList.spec.tsx
@@ -1,7 +1,7 @@
 import { render, screen } from '@testing-library/react';
 import * as React from 'react';
 import { SelectList } from './SelectList';
-import { SemanticColors } from '../../essentials';
+import { getSemanticValue } from '../../utils/cssVariables';
 
 describe('SelectList', () => {
     it('renders options in multi select', () => {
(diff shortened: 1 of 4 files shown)


nlopin:
Those two components have semantic colors in the next branch (see [Tag](https://github.com/freenowtech/wave/blob/258d8ed5415c943bd94eace78615b92b660a101e/src/components/Tag/Tag.tsx#L43C1-L43C1)). In the main, error states were added.

Have you used the same semantic values as in the next branch? That way, we can avoid conflicts

nlopin (approved):

martimalek:
The force pushes that you see are because I corrected the git history of #380, when merging `main` into it I renamed the merge commit by mistake, which made me lose the commits history between `main` and `next`. Now that it is corrected there I'm just updating it here.

martimalek:
> Those two components have semantic colors in the next branch (see [Tag](https://github.com/freenowtech/wave/blob/258d8ed5415c943bd94eace78615b92b660a101e/src/components/Tag/Tag.tsx#L43C1-L43C1)). In the main, error states were added.
> 
> Have you used the same semantic values as in the next branch? That way, we can avoid conflicts

Yes, I've used the same semantic colors that were already in `next`, I've only added the ones that were missing for the new variants

PR #385, opened 2023-10-03
Add dark mode

nlopin:
**What:**

Adds dark mode to both themes: Classic and Redesigned

​
**Why:**

We want our users to enjoy the darkness
​
**How:**

* I added a media query that redefines the theme CSS variables
* `Text`, `Header` and other components now use `color` set to `inherit` to avoid manual overwriting
* Page color now explicitly set in global styles

**What's left**

- [x] Review and verify the color mode component. Find a better name
- [x] Add new logo 
- [x] make the dark schemes satisfy the `Color` type

**Checklist:**

-   [x] Ready to be merged


src/codemods/__testfixtures__/block-to-width-100.output.tsx
```diff
@@ -2,7 +2,7 @@ import { Button } from '@freenow/wave';
 import React from 'react';
```
martimalek, line 2: It seems prettier is running against codemods fixtures and formatting them, this should not happen and we need to rollback these files (everything under /codemods/__testfixtures_). I've created PR #387 so that prettier ignores codemods fixtures from now on

src/components/ColorMode/ColorMode.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const renderColorMode = ({ dark }: { dark: boolean }) =>
+    render(
+        <ColorMode dark={dark}>
+            <Button>Click</Button>
+        </ColorMode>
+    );
+
+describe('ColorMode', () => {
+    it('renders a div wrapper with light-mode by default', () => {
```
martimalek, line 14: Why does it say "by default" when we are explicitly passing `dark: false`?

```suggestion
    it('renders a div wrapper with light-mode if we pass dark prop as false', () => {
```

src/components/Headline/docs/Headline.storybook.mdx
```diff
@@ -8,7 +8,7 @@ import * as HeadlineStories from './Headline.stories';
 
 # Headline
 
-The Headline component will render an html `h1-6` tag. Please reference the
+The Headline component will render a html `h1-6` tag. Please reference the
```
martimalek, line 11: Shouldn't `an` be used before `html` since the `h` is silent, so the first sound you pronounce is the vowel `a`?

martimalek, line 11: Shouldn't `an` be used before `html` since the `h` is silent, so the first sound you pronounce is the vowel `a`?
Sorry for the extreme nitpicking 😅, I'm pointing it out since this is user facing, I had to [google it](https://www.scribbr.com/commonly-confused-words/a-vs-an/) to be sure hahaha 
```suggestion
The Headline component will render an html `h1-6` tag. Please reference the
```

src/components/index.ts
```diff
@@ -57,3 +57,4 @@ export { Row, RowProps, Column, ColumnProps } from './Grid/Grid';
 export { Popover, PopoverProps } from './Popover/Popover';
 export { InfoBanner, InfoBannerProps, InfoBannerVariants } from './InfoBanner/InfoBanner';
 export { InfoBannerCard } from './InfoBanner/blocks/InfoBannerCard';
+export { ColorMode } from './ColorMode/ColorMode';
```
martimalek, line 60: Shouldn't we export `InvertedColorMode` as well?

```suggestion
export { ColorMode } from './ColorMode/ColorMode';
export { InvertedColorMode } from './ColorMode/InvertedColorMode';
```

martimalek (approved):
Awesome!! We are getting really close to Wave 2.0! 🥳 🍾

src/essentials/Colors/Colors.ts
```diff
@@ -199,3 +212,151 @@ export const SemanticColors = {
...
+            faded: Colors.orange[900],
+            default: Colors.orange[900]
+        },
+        focus: Colors.blue.secondary[900],
+        disabled: Colors.blue.primary[550]
+    },
+    shadow: {
+        default: Colors.blue.primary[550]
+    }
+}; // satisfies SemanticColorsSchema;
```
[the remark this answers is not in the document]
nlopin: This will be uncommented

src/components/index.ts
nlopin: Nice catch. This is the part I need to finalize

src/components/Headline/docs/Headline.storybook.mdx
nlopin: I think I just followed my spellchecker here

src/essentials/Colors/Colors.ts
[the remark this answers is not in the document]
nlopin: this is an ongoing thing, added to the description above

src/components/ColorScheme/docs/DarkScheme.storybook.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+There are situations where you want to force a specific color scheme to a component, or a group of components.
+Wave provides a few components to help you with this.
+
+## Invert Color Scheme
+
+`InvertedColorScheme` component inverts the color scheme of all children components:
+
+-   _dark_ scheme for users with light scheme
+-   _light_ scheme for users with light scheme
```
martimalek, line 21:
```suggestion
-   _light_ scheme for users with dark scheme
```

PR #386, opened 2023-10-03
fix(deps): override @types/react from react-windowed-select

### LINKED ISSUE #362: Password and Input components require `crossOrigin` prop
`@freenow/wave` version: latest

### Relevant code

When the `Password` component is used with react types:

"@types/react": "18.2.21",
"@types/react-dom": "18.2.7"

It requires `crossOrigin` property to be set 

the property is not relevant to inputs, and it shouldn't be required ([mdn](http://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#crossorigin))

### Reproduction

[Code Sandbox](https://codesandbox.io/s/serene-beaver-fn8ds4?file=/src/App.tsx)

martimalek:
**What:**

Override the `@types/react` sub-dependency of `react-windowed-select` so that it's always at least version `17.0.67`.
Apart from that, I've also moved `@testing-library/react-hooks` from `dependencies` to `devDependencies`, this is a testing dependency and should not be needed when installing Wave in a project.

​
**Why:**

The existing version of the `@types/react` sub-dependency of `react-windowed-select` in our `package-lock.json` was `17.0.53`, which has a `crossOrigin` prop in the type `InputHTMLAttributes`, which is used as part of the `Input` and `Password` components props. That `crossOrigin` prop was creating an issue which made every project using the latest Wave version have to add it to their `Input`/`Password` components.

​
**How:**

- Add `"@types/react": "^17.0.67"` as overrides of `react-windowed-select`
- Move `@testing-library/react-hooks` to devDependencies

​
**Media:**

Screenshot of our current (in `main`) `package-lock.json` to show that `@types/react` `17.0.53` is installed:
image

**Checklist:**

-   [x] Ready to be merged

Closes #362


nlopin:
How did you verify it works?

nlopin (approved):

package.json
```diff
@@ -48,7 +48,10 @@
     "license": "Apache-2.0",
     "overrides": {
         "react-refresh": "0.11.0",
-        "react-is": "17.0.2"
+        "react-is": "17.0.2",
+        "react-windowed-select": {
+            "@types/react": "^17.0.67"
```
arturmiglio, line 53: Shouldn't this be a fixed version?

```suggestion
            "@types/react": "17.0.67"
```

arturmiglio (approved):

PR #390, opened 2023-10-10
fix: enhanced breadcrumbs style

div-Leo:
###  **What:**
enhance breadcrumbs style:
- nowrap when is longer than the container
- accept margin props
- scroll right if it's longer than the container
​
###  **Why:**
The design was not ideal breaking every word and going to a new line.
​
###  **How:**
- adjust the style of the `ul` and `li`s
- use the ref to change the scroll
​
### **Media:**

**Before** 
Screenshot 2023-10-10 at 16 48 17

**After** 
<video width="347" alt="Screenshot 2023-10-10 at 16 48 17" src="https://github.com/freenowtech/wave/assets/12762609/553615a5-98f4-4fc4-9106-e7e7f1c44623">


JanHamara:
Looks okay 👍

PR #396, opened 2023-11-08
Colors codemod types

### LINKED ISSUE #394: Colors to css variables codemod does not transform types
### Relevant code

When running the `colors-to-css-vars` codemod it is not transforming the usages of `Colors` as a type.

Before
```js
import { Colors } from '@freenow/wave'

interface LabelVariant {
    color?: Colors
    backgroundColor?: Colors
    borderColor?: Colors
}
```
After
```js
interface LabelVariant {
    color?: Colors
    backgroundColor?: Colors
    borderColor?: Colors
}
```

The codemod should transform the `Colors` type to `ReadCssColorVariable`.

After should look like
```js
interface LabelVariant {
    color?: ReadCssColorVariable
    backgroundColor?: ReadCssColorVariable
    borderColor?: ReadCssColorVariable
}
```

martimalek:
**What:**

Make the colors-to-css-vars codemod take into account `Colors` being used as a type instead of as a constant
​
**Why:**

When `Colors` was used as a type the codemod was not transforming it, which made the resulting file invalid

```tsx
import { Colors } from '@freenow/wave';

export interface AnotherInterface { color?: Colors }
```
should be transformed to
```tsx
import { ReadCssColorVariable } from '@freenow/wave';

// eslint-disable-next-line @typescript-eslint/ban-types you can remove this comment and only use the `ReadCssColorVariable` type after changing the bare colors to `getSemanticValue`
export interface AnotherInterface { color?: ReadCssColorVariable | (string & {}) }
```
​
**How:**

- In `colors-to-css-vars` codemod, find usages of `Colors` as a type, replace them for the `ReadCssColorVariable` type and add it as an import
- I've also added an early return checking if `Colors` is imported, so that in case it isn't we stop executing the codemod
- Since the codemod adds an import for `ReadCssColorVariable`, I've exported the type from a `essentials/index.ts`

**Checklist:**

- [x] Add eslint-disable comment to Colors type replacement so it doesn't break linting
- [x] Ready to be merged

Closes #394


src/essentials/index.ts
```diff
@@ -1,3 +1,4 @@
 export { Elevation } from './Elevation/Elevation';
 export { Spaces } from './Spaces/Spaces';
 export { Breakpoints, MediaQueries } from './Breakpoints/Breakpoints';
+export type { ReadCssColorVariable } from './Colors/types';
```
nlopin, line 4: Let's also expose all wave css variables
```suggestion
export type { ReadCssColorVariable, ReadCssVariable } from './Colors/types';
```

nlopin (approved):

src/essentials/index.ts
martimalek: I've ended up creating a `BareColorsSchema` following the `SemanticColorsSchema` and appropriate types based on that, please take a look again when you have time 🙏🏽

src/essentials/Colors/types.ts
```diff
@@ -4,6 +4,58 @@ export type HSL = `hsl(${number}, ${number}%, ${number}%)` | `hsla(${number}, ${
 
 export type Color = HSL | 'transparent';
 
+export type BareColorsSchema = {
```
nlopin, line 7: Wait, this is incorrect. There is no bare schema. The object differ between different themes. For example, Modern schema has one blue color and two types of red

src/essentials/index.ts
nlopin: I intentionally do not expose bare values because they should only be used as an escape hatch

src/essentials/Colors/types.ts
martimalek: Yeah, I understand now, the classic and modern colors structures differ, so we can't have a common schema. I'm thinking on what the best approach is then, because I need a replacement type for what was before `Colors` and is now the bare colors css variables (i.e. `'var(--wave-b-color-white)' | 'var(--wave-b-color-black)' | 'var(--wave-b-color-blue-primary-1100)' | ...`)

I could rename `BareColorsSchema` to `ClassicBareColorsSchema` to make it clear that it's not a common schema for both modern and classic (for modern I don't need it at all). Wdyt? Do you see other possible solutions?

src/essentials/index.ts
martimalek: I understand and agree with the intention, but since projects are currently using `Colors` as a type I need a replacement type for that, I only see as options either exposing a type representing the bare colors as css variables or a generic `string`.

src/codemods/__testfixtures__/colors-to-css-vars/color-as-type.output.tsx
```diff
@@ -0,0 +1,7 @@
+import { ReadBareCssColorVariable } from '@freenow/wave';
+
+export interface LabelVariant {
+    color?: ReadBareCssColorVariable
```
nlopin, line 4:
```suggestion
    color?: ReadCssColorVariable | (string & {})
```

martimalek: Do you think there would be any issue using just `ReadCssColorVariable | string`? I've tested it and it was fine, but maybe I've overlooked something.

The issue with `{}` is that there is an Eslint rule which bans `{}` as a type because it means "any non-nullish value". I'm thinking that if a project has this rules enabled they will need manual changes here 🤔

src/essentials/index.ts
martimalek: I've removed the bare colors schema to not expose bare colors

src/essentials/Colors/types.ts
martimalek: I've removed this in the end

src/codemods/__testfixtures__/colors-to-css-vars/color-as-type.output.tsx
martimalek: When using `ReadCssColorVariable | string` we would lose the intellisense, I've tried it with `ReadSemanticCssColorVariable | (string & NonNullable<unknown>)` instead and we keep the intellisense while avoiding the ESLint rule. `NonNullable<unknown>` works because internally it's just doing `unknown & {}`, as long as we keep the `{}` the intellisense works. Wdyt?

nlopin: I propose to disable the rule for this line (with the reasoning) instead of hacks

martimalek: Even though I like the idea adding comments is not straightforward, different types of nodes need different types of comments, for example if I just add a comment to the existing `TSTypeReference` node the comment will be written before the actual type 😅, instead I need to find the parent node and then attach the comment there, but the parent node can have different types, and attaching a trailing comment works differently for each type.

Just to give an example of the complexity of adding comments I've found [this codemod](https://github.com/amanda-mitchell/suppress-eslint-errors/blob/main/transforms/suppress-eslint-errors.js), which disables eslint rules and adds a comment on top, ours would be a bit simpler, but not by a lot. 

I've also tried adding an `eslint-disable-next-line` comment on the line before the type usage but it also didn't seem straightforward.

Nevertheless I'll take a deeper look next week to see if I can come up with an easy solution that works for all our use cases.

src/codemods/colors-to-css-vars.ts
```diff
@@ -117,6 +120,9 @@ export default (file: FileInfo, api: API, options: Options) => {
     // Find Colors named imports in @freenow/wave imports
     const colorsImports = waveNamedImports.filter(path => path.node.imported.name === 'Colors');
 
+    // Early return in case colors is not used
+    if (colorsImports.size() === 0) return ast.toSource(printOptions);
```
martimalek, line 124: I need to change this to return `file.source` instead, so it does not even transform the code back from ast and returns the original source file instead

martimalek: Already done in #400

martimalek:
I've created the [colors-codemod-add-eslint-comment](https://github.com/freenowtech/wave/tree/colors-codemod-add-eslint-comment) branch with some tries on how to add the eslint disable comment to usages of `Colors` as a type. 

I'll have to park this for a bit due to other priorities and I've created that branch in case we want to move forward with the current solution without the comments.

src/codemods/__testfixtures__/colors-to-css-vars/color-as-type.output.tsx
martimalek: In the end I've been able to add a leading eslint-disable with the reasoning comment each time we replace a usage of `Colors` as a type 👌🏽

nlopin (approved):
Nice! I've added the type cleanup step to the migration guide

PR #403, opened 2024-01-09
Execute all codemodes using one command

### LINKED ISSUE #399: Allow to execute all codemods with one command

Currently we only allow to execute a codemod at a time, we have 6 codemods already which means each developer migrating to Wave 2.0 will need to manually call the 6 codemods one after the other, which can be bothersome (and there are already plans to create a 7th codemod).

A simple approach that we can follow to improve this is to create a "master" codemod that runs all other codemods, this way a developer will be able to only run 1 codemod and automatically complete most of the migration. The option to run each codemod manually will still be available.

This has the limitation that we won't allow the developer to choose which codemods to run, all of them will execute.

> We've decided to go for the "master" codemod approach for simplicity

We need to check what happens when there is an error in one of the codemods.

Alternatively we could create a _cli_ that allows to select which codemods to run, but since I don't have a lot of experience with clis I'm considering creating the "master" codemod.

nlopin:
Closes #399

## What

Allows users to run all codemods with a command `npx @freenow/wave run_v2_migration src/**/*`
​
## How

I made several attempts, but the leaner one is to have a proxy Bash script that passes all provided arguments to the jscodeshift CLI. 

It scans the _codemods_ folder and executes the jscodeshift with every transformation function. If we add more migrations, they will be executed together with old.

​
## Media
Example Run

### DIFF
--- a/docs/migrating.storybook.mdx
+++ b/docs/migrating.storybook.mdx
@@ -20,8 +20,20 @@ allow jscodeshift to parse your typescript/jsx files.
 
 ## From v1 to v2
 
-The following codemods exist to upgrade your components for version 2. In order to apply the transformations in your project you simply need to run the
-command for your desired codemod pointing to your project's source path.
+Run all migrations at once:
+
+```bash
+# npm
+npx @freenow/wave run_v2_migration src/**/*
+
+# yarn
+yarn exec node_modules/.bin/run_v2_migration src/**/*
+
+```
+
+Below, you'll see individual codemods to upgrade components to version 2.
+In order to apply an individual transformation to your project
+you need to run the command for your desired codemod pointing to your project's source path.
 
 ### Button and TextButton block property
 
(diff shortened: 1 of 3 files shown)


div-Leo (approved):

docs/migrating.storybook.mdx
```diff
@@ -20,16 +20,28 @@ allow jscodeshift to parse your typescript/jsx files.
```
martimalek, line 20:
```suggestion

Check out the [other Jscodeshift CLI options](https://github.com/facebook/jscodeshift?tab=readme-ov-file#usage-cli) to further customise the codemods.
```

docs/migrating.storybook.mdx
```diff
@@ -20,16 +20,28 @@ allow jscodeshift to parse your typescript/jsx files.
...
+
+```bash
+# npm
+npx @freenow/wave run_v2_migration src/**/*
+
+# yarn
+yarn exec node_modules/.bin/run_v2_migration src/**/*
+
+```
+
```
martimalek, line 33:
```suggestion

This command also accepts Jscodeshift's CLI options.
```

martimalek (approved):
I've added some suggestions for the docs, I tested the command in partner-dashboard and it executed fine 👌🏽

nlopin:
I've added all the Migration guide suggestions to a separate PR: #406

Issue #404, opened 2024-01-11
CSS Variables are missing on the non-story pages


nlopin:
The Welcome page and all the other components unrelated to components do not have CSS semantic variables injected. 

Wrong styles:
![CleanShot 2024-01-11 at 11 24 30@2x](https://github.com/freenowtech/wave/assets/1469636/9535b1eb-a3c9-43af-9a82-ab4f0d4e38f3)

Correct styles:
![CleanShot 2024-01-11 at 11 25 04@2x](https://github.com/freenowtech/wave/assets/1469636/6bcba98c-3707-4760-8e38-3938ee1f9488)

PR #409, opened 2024-01-19
Docs go dark

### LINKED ISSUE #408: Dark mode for Storybook
We support dark schema in the design system, but not in Storybook. We need to add dark theme for better user experience

nlopin:
## What

Adds dark mode support and a switch between dark and light mode to the documentation. The switch also switches the Wave's scheme.

Resolves #408 #404

### Media

The updated documentation is available on [the temporary website](https://freenowtech.github.io/tmp-wave-v2-docs/).

CleanShot 2024-01-19 at 18 28 01

Video
​
## How

- Storybook docs don't explain theming well. After some research, I found that they do not support theme switches out of the box. I found an [addon `storybook-dark-mode`](https://github.com/hipstersmoothie/storybook-dark-mode) that adds almost everything I needed.
- The `Source` components didn't respect the mode setting. I've written a small wrapper `Source` that derives the theme and passes it to the Storybook `Source` component. As a result, I had to change paths for all the users of the `Source`
- A lot of small adjustments to make the dark mode okay


.storybook/main.ts
```diff
@@ -6,7 +6,8 @@ const config: StorybookConfig = {
         '@storybook/addon-links',
         '@storybook/addon-essentials',
         '@storybook/addon-interactions',
-        '@storybook/addon-viewport'
+        '@storybook/addon-viewport',
+        'storybook-dark-mode'
```
martimalek, line 10: It seems like storybook has their own addon for witching themes ([@storybook/addon-themes](https://storybook.js.org/addons/@storybook/addon-themes)), did you try it before deciding to use this other one? Checking it's docs they are quite lacking, is that what you meant in the description when saying that storybook doesn't explain theming well? 🤣

martimalek (approved):
LGTM, I've just added a comment to make sure we've tried storybook's own addon before using another one, feel free to resolve it if you think it's not worth trying

div-Leo (approved):
Amazing job!

.storybook/main.ts
nlopin: It only allow changing the theme for preview. We are changing the whole documentation theme together with the preview

PR #411, opened 2024-01-29
Semantic colors codemod

### LINKED ISSUE #395: Add codemod for semantic colors

We already have a codemod that replaces usages of `Colors` for css variables. We should try to do the same with `SemanticColors`. 

In this case we'll need to transform usages of `SemanticColors` constant (or type) to the `getSemanticValue` helper (and the `ReadCssColorVariable` type)

The alternative would be to not create the codemod, which would result in more manual effort needed for each migration to Wave 2.0 in a project.

martimalek:
## What

Adds a codemod to automatically migrate usages of `SemanticColors` and `themeGet` to `getSemanticValue`

**`SemanticColors` example**
From 
```tsx
SemanticColors.button.primary.backgroundDisabled
```
to
```tsx
getSemanticValue('background-element-disabled-default')
```

**`themeGet` example**
From
```tsx
themeGet('semanticColors.text.primary')
```
To
```tsx
getSemanticValue('foreground-primary')
```

## Why

To ease the migration effort throughout our codebase

​
## How

- Create `semantic-colors-to-new-theme` codemod
- Dropped `table.zebraStyles.backgroundSecondary` color, it is not used in any project

> Closes #395


nlopin (approved):

src/essentials/Colors/Colors.ts
```diff
@@ -92,7 +92,7 @@ export const SemanticColors = {
             default: Colors.green[900]
         },
         warning: {
-            default: Colors.blue.primary[900]
+            default: Colors.yellow[900]
```
martimalek, line 95: I saw this value was inconsistent with out previous colors and "corrected" it, let me know if the change doesn't make sense 🙏🏽

nlopin: all good, definitely a bug here

PR #420, opened 2024-02-05
docs: fix semantic colors hsl values in colors table

### LINKED ISSUE #419: HSL values in semantic colors table do not match in dark mode
-   `@freenow/wave` version: 2.x.x

When having the docs in dark mode the hsl values are kept as the same ones for the light mode, which means they don't really match with the actual color displayed

### Relevant code
The bug has to be somewhere near [src/essentials/Colors/docs/SemanticColorsTable.tsx](https://github.com/freenowtech/wave/blob/main/src/essentials/Colors/docs/SemanticColorsTable.tsx)

### Reproduction
image

[See it live here](https://wave.free-now.com/?path=/docs/essentials-colors--docs&globals=theme:classic)

martimalek:
## What

- Fixes a bug in our docs where the hsl values in the semantic colors table were not being refreshed when the light/dark theme changed.
- Make the colors table show the current scheme and the inverted scheme colors at the same time

### Media

https://github.com/freenowtech/wave/assets/46452321/be895b10-1495-4b08-8c45-225520405034

## Why

To have the correct hsl values for dark schema colors.

## How

- Point to light/dark schema object based on the colors scheme the user is viewing
- Reorder light/dark schema colors object so the order of the colors matches (so that the table order is consistent for both schemas)
- Add a column in the `CssVariablesTable` for the inverted scheme color
- Remove the `Hex Color` column in `CssVariablesTable`

Closes #419

### DIFF
--- a/src/essentials/Colors/Colors.ts
+++ b/src/essentials/Colors/Colors.ts
@@ -63,27 +63,27 @@ export const SemanticColors = {
     transparent: Colors.transparent,
     foreground: {
         primary: Colors.blue.primary[900],
-        'on-background': {
-            primary: Colors.white,
-            accent: Colors.white,
-            info: Colors.white,
-            success: Colors.white,
-            warning: Colors.blue.primary[900],
-            danger: Colors.white,
-            disabled: Colors.white,
-            neutral: Colors.blue.primary[900]
+        accent: {
+            default: Colors.blue.secondary[900],
+            emphasized: Colors.blue.secondary[1000]
         },
-        disabled: Colors.blue.primary[200],
-        focus: Colors.blue.secondary[900],
         neutral: {
             faded: Colors.blue.primary[200],
             default: Colors.blue.primary[350],
             emphasized: Colors.blue.primary[550]
         },
-        accent: {
-            default: Colors.blue.secondary[900],
-            emphasized: Colors.blue.secondary[1000]
+        'on-background': {
+            primary: Colors.white,
+            accent: Colors.white,
+            neutral: Colors.blue.primary[900],
+            disabled: Colors.white,
+            info: Colors.white,
+            success: Colors.white,
+            warning: Colors.blue.primary[900],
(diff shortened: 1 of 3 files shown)


nlopin (approved):
🥇

src/essentials/Colors/docs/SemanticColorsTable.tsx
```diff
@@ -58,23 +105,40 @@ export const CssVariablesTable: FC<{ tier: 'b' | 's' }> = ({ tier }) => {
...
+                        <TableRow key={variable}>
+                            <TableCell>
+                                <BlockContainer>
+                                    <ColorBlock token={token} />
+                                </BlockContainer>
+                                <code>{value}</code>
+                            </TableCell>
+                            <TableCell>
+                                <BlockContainer>
+                                    {isDark ? (
```
martimalek, line 124: I tried using `InvertedColorScheme` but the color would always be the same, I don't understand why, will try debugging this more tomorrow

nlopin: It is expected, the `InvertedColorScheme` relies on your OS preference, not on Storybook toggle

nlopin (approved):

src/essentials/Colors/Colors.ts
```diff
@@ -63,27 +63,27 @@ export const SemanticColors = {
...
+        },
         'on-background': {
             primary: Colors.white,
             accent: Colors.white,
+            neutral: Colors.blue.primary[900],
+            disabled: Colors.white,
             info: Colors.white,
             success: Colors.white,
             warning: Colors.blue.primary[900],
-            danger: Colors.white,
```
nlopin, line 72: Why are we deleting tokens?

martimalek: I'm not deleting them, just moving them around so all have the same order (in order for the tables in each scheme to match)

PR #425, opened 2024-02-08
Fix datepicker elevation, props and bottom-lined input ref

### LINKED ISSUE #421: Datepicker appears underneath Modal
-   `@freenow/wave` version: 2.1.0 (older versions as well, in the v1 docs it also happens)

### Relevant code
Datepicker appears underneath the Modal when adding it inside it

Code taken from `DatepickerOnModal` story
```tsx
<>
    <Button size="small" onClick={() => setShowModal(true)}>
        Open Modal with Datepicker
    </Button>
    {showModal && (
        <Modal dismissible={false} onClose={() => setShowModal(false)}>
            {dismiss => (
                <>
                    <Headline as="h2">New Event</Headline>

                    <DatePicker value={value} onChange={setValue} />

                    <br />
                    <Button onClick={dismiss}>Add Event</Button>
                    <TextButton onClick={dismiss}>Cancel</TextButton>
                </>
            )}
        </Modal>
    )}
</>
```

Screenshot of the above code in our docs
image

Datepicker should appear above the Modal, not underneath it

### Reproduction

Check the issue [in our docs site](https://wave.free-now.com/?path=/docs/form-elements-datepicker-datepicker--docs)

martimalek:
## What

- Fixes Datepicker elevation when appearing on a Modal
- Fixes Datepicker props not extending Input props
- Fixes "bottom-lined" Input variant ref

### Media
Before (Datepicker underneath Modal)
image

After (Datepicker on top of Modal)
image

## Why

To offer a consistent experience on the Datepicker and Input components

## How

- Move z-index to the highest Datepicker container (before it was in a lower level container and the Modal would take preference)
- Extend `InputProps` in `DatepickerSingleInputProps` since props are being sent to trigger `Input`
- Point ref in `InnerInput` with "bottom-lined" variant to the actual html `input` instead of the the wrapper `div`
- Minor addition of a label for the "Datepicker on Modal" story

Closes #421

### DIFF
--- a/src/components/Datepicker/Datepicker.tsx
+++ b/src/components/Datepicker/Datepicker.tsx
@@ -3,7 +3,7 @@ import { useDatepicker, MonthType, UseDatepickerProps } from '@datepicker-react/
 import styled from 'styled-components';
 
 import { getSemanticValue } from '../../utils/cssVariables';
-import { Elevation, MediaQueries } from '../../essentials';
+import { MediaQueries } from '../../essentials';
 import { ChevronLeftIcon, ChevronRightIcon } from '../../icons';
 import { Month } from './Month';
 import { DatepickerContext } from './DatepickerContext';
@@ -21,7 +21,6 @@ const DatepickerContainer = styled.div`
     display: flex;
     padding: 0.5rem;
 
-    z-index: ${Elevation.DATEPICKER};
     position: relative;
     background: ${getSemanticValue('background-surface-neutral-default')};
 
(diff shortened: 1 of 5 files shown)


nlopin:
I'd try attaching the datepicker popup to the modal, not to the body. In that case, the elevation of the modal will be used and we can rely on it. I'm not confident it will work, but I don't see the reason why it shouldn't

martimalek:
> I'd try attaching the datepicker popup to the modal, not to the body. In that case, the elevation of the modal will be used and we can rely on it. I'm not confident it will work, but I don't see the reason why it shouldn't

@nlopin how can we know to which element to attach it to? Do you mean having a new prop (e.g. `portalContainer`)?

nlopin:
> > I'd try attaching the datepicker popup to the modal, not to the body. In that case, the elevation of the modal will be used and we can rely on it. I'm not confident it will work, but I don't see the reason why it shouldn't
> 
> @nlopin how can we know to which element to attach it to? Do you mean having a new prop (e.g. `portalContainer`)?

yep, that's the issue. Let's merge this solution and see if the z-index war will start 😂

nlopin (approved):

PR #427, opened 2024-02-08
Fix applied color scheme on portals

### LINKED ISSUE #426: Enforced Color Scheme is not applied on Portals
-   `@freenow/wave` version: 2.x.x

### Relevant code

In `DatepickerSingleInput`, `DatepickerRangeInput` and `Tooltip` we are using [Portals](https://react.dev/reference/react-dom/createPortal) to escape the container element and make sure it doesn't interfere with our components.

These portals will most likely render outside the React root (top level container that you pass to `createRoot(container).render(X)`), in case a color scheme is enforced for the root it won't be applied to the portal.

Enforced color schemes apply to our components rendered in portals.

### Reproduction

Datepicker container has the light scheme enforced, but the Datepicker appears with dark colors
image

Check it out in [this codesandbox](https://codesandbox.io/p/sandbox/wave-v2-color-scheme-on-portals-3pzkpy)

martimalek:
## What

Fixes a bug where enforced color schemes would not get applied to components in Portals (Datepickers and Tooltip)

### Media

Light scheme is enforced to Datepicker but only gets applied to the trigger input, the actual Datepicker is still dark
image

Corrected behaviour
 
image

## Why

If a user wants to enforce a color scheme to a component it should get applied to the whole component, not just part of it

## How

- Create a utility hook `useClosestColorScheme` that finds the enforced color scheme closest to the passed element
- Wrap portal content in `LigthScheme`/`DarkScheme`/`React.Fragment` based on the result of `useClosestColorScheme` passing the trigger element (e.g. the input that renders the Datepicker when you click on it), this is done for both Datepickers as well as Tooltip (all components that use portals)

Closes #426

### DIFF
--- a/src/utils/hooks/useClosestColorScheme.ts
+++ b/src/utils/hooks/useClosestColorScheme.ts
@@ -1,0 +1,20 @@
+import { useMemo } from 'react';
+
+/**
+ * useClosestColorScheme - Returns 'light' / 'dark' based on the enforced color scheme closest to the element passed or `undefined` when no color scheme is enforced
+ */
+export const useClosestColorScheme = (element?: HTMLElement): 'light' | 'dark' | undefined =>
+    useMemo(() => {
+        if (!element) return undefined;
+
+        const closestParentWithColorScheme = element.closest('.wave');
+        if (!closestParentWithColorScheme) return undefined;
+
+        const hasEnforcedLightScheme = closestParentWithColorScheme.classList.contains('light-scheme');
+        if (hasEnforcedLightScheme) return 'light';
+
+        const hasEnforcedDarkScheme = closestParentWithColorScheme.classList.contains('dark-scheme');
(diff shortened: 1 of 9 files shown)


nlopin (approved):
Great work, Marti! Elegant idea

src/utils/hooks/useClosestColorScheme.ts
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from 'react';
+
+/**
+ * useClosestColorScheme - Returns 'light' / 'dark' based on the enforced color scheme closest to the element passed or `undefined` when no color scheme is enforced
+ */
+export const useClosestColorScheme = (element?: HTMLElement): 'light' | 'dark' | undefined =>
```
nlopin, line 6: Should we also subscribe for changes? In the case of the inverted colour mode, the class will be updated by the browser when OS changes the mode

nlopin (approved):

PR #430, opened 2024-02-13
fix(offcanvas): export offcanvas hook

martimalek:
## What

Fixes an issue where the `useOffcanvasDismiss` hook was not being exported. Fixes issues with scrolling in the docs

### Media

Check out in our docs how scrolling is not possible for the [Offcanvas](https://wave.free-now.com/?path=/docs/components-offcanvas--docs) and [Modal](http://localhost:6006/?path=/docs/components-modal--docs) docs
​
Modal scrolling now

https://github.com/freenowtech/wave/assets/46452321/03475132-00e9-45df-baaa-ea0eb79e87d3

Offcanvas scrolling now

https://github.com/freenowtech/wave/assets/46452321/00e4e291-bbea-41d3-b03f-30b43f59b2fd

## Why

We need to export `useOffcanvasDismiss` to be able to dismiss the `Offcanvas` through a hook
We should be able to scroll in our docs site 😅 
​
## How

- Export `useOffcanvasDismiss` in the components index
- Removed `defaultProps` in `Offcanvas` in favour of default arguments in the component
- Remove `Modal` and `Offcanvas` stories that use a `render` other than the default, having multiple stories using `render` causes this weird issue where scrolling stops working

### Extra ❗ 

In my opinion we can safely remove those stories since all of them were to demo the render props vs hook dismiss options and they are already explained in a section on top of the stories for both Modal and Offcanvas, if you think it's worth keeping them I can open a separate issue for that and keep this PR for only exporting the hook and refactoring the default props

### DIFF
--- a/src/components/Modal/docs/Modal.stories.tsx
+++ b/src/components/Modal/docs/Modal.stories.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { StoryObj, Meta } from '@storybook/react';
-import { Modal, useModalDismiss } from '../Modal';
+import { Modal } from '../Modal';
 import { Button } from '../../Button/Button';
 
 const meta: Meta = {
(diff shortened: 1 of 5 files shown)


nlopin (approved):
I thought it was a Storybook thing. Great work!

Issue #434, opened 2024-03-01
<PhoneInput listPortalTarget={document.body}> doesn't use LightScheme


jahe:
The country code selection menu of `<PhoneInput listPortalTarget={document.body}>` appears with dark scheme colors although `<LightScheme>` is used.

In the documentation it's not reproducible because `light-scheme` is set on `<body>`. In our case `light-scheme` is set on the root `<div>`. Problem is that the selection menu of `<PhoneInput>` is being rendered right under `<body>`. 

-   `@freenow/wave` version: 2.1.2

### Relevant code

```js
const container = document.querySelector('#root');
const root = createRoot(container!);
root.render(
    <>
        <ClassicColors />
        <LightScheme>
            <Root />
        </LightScheme>
    </>
);
```

```js
<PhoneInput listPortalTarget={document.body} />
```

### What was expected to happen?
Country code selection menu appears in light scheme colors.

### Reproduction
https://codesandbox.io/p/sandbox/wave-playground-forked-lxtsvk

martimalek:
Hey @jahe, thanks for opening the issue! I've marked it as a bug 🐛 

If my understanding is correct what's happening is that `PhoneInput` uses `SelectList` underneath and when passing `listPortalTarget` to it you change where in the DOM the list will be rendered, in your case the `<body>`, since it is higher in the DOM than the `LightScheme` it won't have the scheme applied.

The solution will probably be to wrap the list in `LightScheme`/`DarkScheme` depending on the applied scheme on the component. We already have an internal hook that detects the applied scheme and even reacts to changes to it, so this part should be straightforward, the tricky bit will be how to wrap it because we are using `react-windowed-select` to render the list, so we are not directly creating the portal, we'd need to take a look at the code and check how flexible their API is, I don't recall exactly but I think we are able to pass our custom components to the library so that it renders them efficiently, maybe wrapping those components would do the trick.

I can try to take a look at it next week 👀

JanHamara:
@jahe we will merge in the fix today, please in your use case use `menuPortalTarget` on PhoneInput instead of `listPortalTarget` 🙂

martimalek:
Resolved by @JanHamara in #438

PR #438, opened 2024-04-05
fix(SelectList): overwrite styling with enforced color scheme for menu

JanHamara:
## What

Fixing the problem from Issue #434 

### Media

https://www.loom.com/share/a76665438d7745adb3abeb1cf962e7b6?sid=73419f0a-b715-4d97-afa9-e3fc575b82d3

## How

- Use `useClosestColorScheme` to identify if enforced color scheme is applied
- Wrap the `Menu` component from `react-windowed-select` with respective color scheme component

### DIFF
--- a/src/components/SelectList/SelectList.tsx
+++ b/src/components/SelectList/SelectList.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { FC, ReactNode } from 'react';
 import {
     ClearIndicatorProps,
     components as ReactSelectComponents,
@@ -21,6 +21,7 @@ import { Label } from './components/Label';
 import { Wrapper } from './components/Wrapper';
 import { disabledStyles, errorStyles, variantStyles } from './styles';
 import { SelectListProps } from './types';
+import { MenuProps } from 'react-select';
 
 type WithSelectProps<T> = T & { selectProps: SelectListProps };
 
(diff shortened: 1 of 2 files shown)


src/components/SelectList/SelectList.tsx
```diff
@@ -309,21 +311,54 @@ const MultiValueRemove = props => (
     </ReactSelectComponents.MultiValueRemove>
 );
 
+// eslint-disable-next-line react/prop-types
```
martimalek, line 314: Doesn't `react-select` export props for the `Menu` component? If they exist we could get rid of both the eslint and the ts comment

martimalek (approved):
Nice job! 🚀 

I've just left one comment to see if we can type some props and get rid of some warnings... 🤔

src/components/SelectList/SelectList.tsx
JanHamara: good point 👍

src/components/SelectList/SelectList.tsx
```diff
@@ -309,21 +312,48 @@ const MultiValueRemove = props => (
     </ReactSelectComponents.MultiValueRemove>
 );
 
+const LightSchemeMenu = (children, props: WithSelectProps<MenuProps>) => (
+    <LightScheme>
+        <ReactSelectComponents.Menu {...props}>{children}</ReactSelectComponents.Menu>
+    </LightScheme>
+);
+
```
martimalek, line 320: Is there a need to have children as a first argument? I believe it should be a regular prop 🤔 

```suggestion
const LightSchemeMenu = (props: WithSelectProps<MenuProps>) => (
    <LightScheme>
        <ReactSelectComponents.Menu {...props}>{props.children}</ReactSelectComponents.Menu>
    </LightScheme>
);

```

PR #439, opened 2024-04-08
fix: extends InputProps on search and spread into search SyledInput

dornelasnelson:
## What

Make `SearchProps` extend the `InputProps` and spread its props into the `StyledInput` of the `Search` component.

### Media

No UI change.

## Why

Search missed a few `InputProps` to it's `StyledInput`; the one missing that we currently wanted to use is `autoFocus` so that we can have the user automatically focus in the SearchInput with a single attribute.

## How

Making `SearchProps extend InputProps` (omitting results because of conflict) and removing a few props from `SearchProps` (disabled, placeholder, size) as they are already in `InputProps`


martimalek (approved):
Nice one! Thanks for contributing 🤗

PR #442, opened 2024-04-15
fix: remove width from pageSize selectList box

### LINKED ISSUE #331: Pagination page size select too narrow
-   `@freenow/wave` version: 1.30.0

### Relevant code

```js
<Pagination
      pageSize={10}
      pageSizes={[
        { label: "10", value: "10" },
        { label: "20", value: "20" },
        { label: "50", value: "50" },
        { label: "100", value: "100" }
      ]}
    />
```

Current value to be fully visible instead of seeing the first character only

CleanShot 2024-02-07 at 15 05 38@2x

See it in [the docs](https://wave.free-now.com/?path=/docs/components-pagination--docs)

### Reproduction
https://codesandbox.io/s/blue-silence-k1wvpr?file=/src/_app.js:84-303

dornelasnelson:
## What

Fixes the size of the Page Size selector by removing the width attribute sent to it.

### Media

Page size selector with not broken size:
image

Responsiveness of boxes:

https://github.com/freenowtech/wave/assets/153822795/bbe228e3-1486-4cf0-b930-c01bd44a4d45

## Why

There is a bug where the page size selector is too small and does not fully display the number of the selected page size. By removing the set `width="4.5em"` it displays the selected page size correctly.

## How

Adds 3 boxes inside of ButtonsContainers, with the left (PageSize selector) and right (empty) having `flex='1'` so that they take the necessary space, and having the box with the pagination navigation in the middle.

Sets the container inside of the first flexable box with `minWidth='5em'` and `width="fit-content"` so that the PageSize selector can't be as small as it was before, but also the box around it keeps the having the same size as the SelectList 

Closes #331

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -12,8 +12,8 @@ const Container = styled.div`
 
 const ButtonsContainer = styled(Box)`
     display: flex;
-    justify-content: center;
-    position: relative;
+    align-items: center;
+    gap: ${Spaces[1]};
 `;
 
 const IconButton = styled.button`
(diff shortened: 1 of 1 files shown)


dornelasnelson:
> Have you considered setting `minWidth` here to restrict the minimum size?

hey @nlopin, I didn't do that since the Box position is absolute its size doesn't really seem to change with different screen sizes 🤔 

I just tried with '5em' instead of the '4.5em' that we had before and it was too small, it doesn't seem to affect the size at all

nlopin:
I don't understand why we set position to absolute for this element. I suspect it could be because the popup is cut. Could you check that? if we can get rid of the absolute positioning and use flexbox to properly place the element

dornelasnelson:
sure, let me try to do that. I believe the absolute value for the pagesize is there to set it to left while having the pagination navigation buttons right in the middle, but let me try to check what can be done with flex

src/components/Pagination/Pagination.tsx
```diff
@@ -140,34 +140,39 @@ const Pagination: React.FC<PaginationProps> = ({
...
-                    </Box>
-                )}
-                {size !== 'small' && (
-                    <IconButton aria-label={ariaLabelFirst} disabled={isFirstPage} onClick={onSkipBackward}>
-                        <BackwardLastIcon size="small" color="inherit" />
-                    </IconButton>
-                )}
+                <Box flex="1">
+                    {hasMultiplePageSizes && (
+                        <Box aria-label={ariaLabelSelectPageSizeContainer} minWidth="5em" maxWidth="5.5em">
```
nlopin, line 145: shouldn't we remove the `maxWidth`?

dornelasnelson: I edited the description to explain why if we remove the maxWidth the inner box will grow together with the outer one, having the PageSize selector grow way too much:

image

dornelasnelson: As we spoke in Slack, I changed it to have the `width="fit-content"` instead of the maxWidth ;)

nlopin (approved):

martimalek (approved):
Yay! This has been bothering us for some time! Thanks for fixing it 🤗

Issue #448, opened 2024-05-21
Create Semantic Colors For Charts


jahe:
### Semantic Colors For Charts
As discussed internally in this [thread](https://mytaxi.slack.com/archives/CLM50RECU/p1715615753074619) we want to add semantic colors for charts. You can find the design for it [here](https://www.figma.com/design/ctaHZoIjoQkSYPqyD3w1x9/Wave-2.0-%5BWIP%5D?node-id=6578-47&t=U8PibwZr6ggHxJWZ-0).

![image](https://github.com/freenowtech/wave/assets/2736136/ad2ecc01-0e24-4ef3-abfa-969166b5af9c)
![image](https://github.com/freenowtech/wave/assets/2736136/2d957638-5482-4823-b21e-71a2f3658a3a)

I'm gonna take care of adding the colors to Wave.

PR #449, opened 2024-05-21
feat: add semantic colors for chart (#448)

jahe:
## What

Add new semantic colors for charts.

(+ fix few existing primary colors in modern look)

### Media
Modern
image
image

Classic
image
image

## Why
We need those colors in Admin Panel for a histogram component.

​

## How
Add new primary colors and introduce new semantic colors for `chart`.

### DIFF
--- a/src/essentials/Colors/ModernColors.ts
+++ b/src/essentials/Colors/ModernColors.ts
@@ -15,12 +15,17 @@ export const Colors = {
         50: 'hsl(0, 2%, 91%)'
     },
     primary: {
-        1000: 'hsl(346, 51%, 33%)',
+        1100: 'hsl(341, 100%, 13%)',
+        1000: 'hsl(343, 70%, 22%)',
+        950: 'hsl(346, 51%, 33%)',
         900: 'hsl(347, 42%, 43%)',
-        350: 'hsl(349, 89%, 76%)',
-        150: 'hsl(351, 100%, 85%)',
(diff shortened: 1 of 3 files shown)


src/essentials/Colors/ModernColors.ts
```diff
@@ -15,12 +15,17 @@ export const Colors = {
...
-        350: 'hsl(349, 89%, 76%)',
-        150: 'hsl(351, 100%, 85%)',
-        100: 'hsl(352, 100%, 93%)',
-        50: 'hsl(9, 100%, 96%)'
+        500: 'hsl(347, 41%, 50%)',
+        350: 'hsl(349, 50%, 65%)',
+        150: 'hsl(348, 50%, 76%)',
+        120: 'hsl(351, 51%, 85%)',
+        100: 'hsl(352, 48%, 91%)',
+        50: 'hsl(350, 46%, 95%)',
```
martimalek, line 27: For projects already using these colors this will result in breaking changes, I understand this is aligned with design and it's expected, but I'd still communicate this on the internal slack #dev-frontend to make sure everyone is aware

martimalek (approved):

src/essentials/Colors/ModernColors.ts
jahe: Thanks for the hint @martimalek! Will do!

PR #455, opened 2024-06-06
docs: add documentation about distribution channels

martimalek:
## What

Adds documentation around our distribution channels

### Media

image

## Why

So contributors know in which branch to contribute depending on how they want to release the new version


JanHamara (approved):
Thank you for adding this 👍

PR #457, opened 2024-06-13
feat: add two new neutral colors and map them to new semantic color

jahe:
## What
Our designer asked us to add two new neutral base colors to the modern theme and add new semantic colors while deprecating others.

### Media
image

image

image

## Why
We are redefining the background and surface tokens and will need these extra tones. (neutral-10, neutral-800)

## How
- Add new colors and correct the existing neutral colors (some they were slightly off but it's not significant so should't be considered a breaking change).
- Add new semantic color `background-page-elevation-0`
- Change color for `background-page-default` slightly

### DIFF
--- a/src/essentials/Colors/Colors.ts
+++ b/src/essentials/Colors/Colors.ts
@@ -111,9 +111,9 @@ export const SemanticColors = {
         }
     },
     background: {
-        default: Colors.neutral[10],
         page: {
             default: Colors.white,
+            'elevation-0': Colors.blue.primary[50],
             'elevation-1': Colors.white,
             'elevation-2': Colors.white,
             'elevation-3': Colors.white
@@ -183,13 +183,6 @@ export const SemanticColors = {
             }
         },
         backdrop: Colors.blue.primary[900]
-    },
-    surface: {
-        default: Colors.white,
-        container: {
-            default: Colors.neutral[10],
-            highest: Colors.white
-        }
     },
     border: {
         accent: {
@@ -284,9 +277,9 @@ export const SemanticColorsDarkSchema = {
         }
     },
     background: {
-        default: Colors.neutral[900],
         page: {
-            default: Colors.blue.primary[900],
+            default: Colors.blue.primary[1000],
+            'elevation-0': Colors.blue.primary[1100],
             'elevation-1': Colors.blue.primary[750],
             'elevation-2': Colors.blue.primary[550],
             'elevation-3': Colors.blue.primary[350]
@@ -356,13 +349,6 @@ export const SemanticColorsDarkSchema = {
             }
         },
         backdrop: Colors.blue.primary[50]
-    },
-    surface: {
-        default: Colors.neutral[800],
-        container: {
-            default: Colors.neutral[750],
-            highest: Colors.neutral[650]
-        }
     },
     border: {
         accent: {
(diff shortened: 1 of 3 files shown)


arturmiglio (approved):

src/essentials/Colors/Colors.ts
```diff
@@ -265,6 +284,7 @@ export const SemanticColorsDarkSchema = {
         }
     },
     background: {
+        default: Colors.neutral[900],
         page: {
             default: Colors.blue.primary[900],
```
arturmiglio, line 289: @jahe as agreed with Anna on Friday, we will only introduce a new semantic key: `background-page-elevation-0`.

It should be mapped to the new `neutral-10`. The rest of the semantic names/values should be untouched.

Thank you!

src/essentials/Colors/types.ts
```diff
@@ -9,11 +9,12 @@ export type SemanticColorsSchema = {
...
+        default: Color;
         page: {
-            default: Color;
-            'elevation-1': Color;
-            'elevation-2': Color;
-            'elevation-3': Color;
+            default: Color; // deprecated - use background.default or surface.default instead
+            'elevation-1': Color; // deprecated - use surface.container.default instead
+            'elevation-2': Color; // deprecated - use surface.container.highest instead
+            'elevation-3': Color; // deprecated
```
arturmiglio, line 17: We won't deprecate those for now.

arturmiglio (changes requested):

src/essentials/Colors/Colors.ts
```diff
@@ -5,6 +5,17 @@ import { createThemeGlobalStyle } from './globalStyles';
 export const Colors = {
     white: 'hsl(0, 0%, 100%)',
     black: 'hsl(0, 0%, 0%)',
+    neutral: {
```
martimalek, line 8: These are the "classic" (old) bare colors, I'm not sure if these neutrals make sense here, I understood they belong to the new palette only 🤔

src/essentials/Colors/types.ts
```diff
@@ -9,11 +9,12 @@ export type SemanticColorsSchema = {
     white: Color;
     black: Color;
     background: {
+        default: Color;
         page: {
-            default: Color;
-            'elevation-1': Color;
-            'elevation-2': Color;
-            'elevation-3': Color;
+            default: Color; // deprecated - use background.default or surface.default instead
```
martimalek, line 14: In alignment with Artur's comments, I also understood after the call with design that the only changes to our semantic tokens would be adding `elevation-0`

martimalek (changes requested):

src/essentials/Colors/Colors.ts
jahe: Hey @arturmiglio! I applied the changes we agreed on in 86d8c03.

src/essentials/Colors/types.ts
jahe: fixed in 86d8c03

src/essentials/Colors/types.ts
jahe: Yup, you're right Marti. I applied the changes in 86d8c03.

src/essentials/Colors/Colors.ts
jahe: Hey @martimalek! Good catch!

In one of the previous iterations this was the suggestion from design that we introduce `neutral` colors to the old classic theme. However now that we agreed to not use any of them I removed them again in 6f84898.

arturmiglio (approved):

martimalek (approved):

PR #462, opened 2024-07-23
fix(Pagination): Stop clipping two digit page sizes with fixed width

JanHamara:
## What
Removes fixed width of Pagination page size selector, to avoid clipping the page size number

### Media

​_Before_
Screenshot 2024-07-23 at 14 52 55

​_After_
Screenshot 2024-07-23 at 14 56 28

## Why
We had reports from DMT that page size is not visible with 2+ digits number.

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -135,7 +135,7 @@ const Pagination: React.FC<PaginationProps> = ({
         <Container>
             <ButtonsContainer>
                 {hasMultiplePageSizes && (
-                    <Box aria-label={ariaLabelSelectPageSizeContainer} position="absolute" left="0" width="4.5em">
+                    <Box aria-label={ariaLabelSelectPageSizeContainer} position="absolute" left="0">
                         <SelectList
                             options={pageSizes}
                             onChange={onSelectPageSize}


src/components/Pagination/Pagination.tsx
```diff
@@ -135,7 +135,7 @@ const Pagination: React.FC<PaginationProps> = ({
         <Container>
             <ButtonsContainer>
                 {hasMultiplePageSizes && (
-                    <Box aria-label={ariaLabelSelectPageSizeContainer} position="absolute" left="0" width="4.5em">
+                    <Box aria-label={ariaLabelSelectPageSizeContainer} position="absolute" left="0">
```
arturmiglio, line 138: Not directly related, but I think the width here was an atempt to minimise a possible collision between the nav buttons and the and the dropdown.

Screenshot 2024-07-23 at 16 09 51 
(taken from the running the artifact of storybook locally)

Not sure what/if something changed recently but this is not happening currently in one implementation (at least):

Screenshot 2024-07-23 at 16 12 15

Do you think you can have a look at how we place the page size selector in the container so it has block size and make the other buttons wrap, instead of overlapping with them? 

Thanks in advance!

arturmiglio: Just realised that this a fix for the V1. So feel free to ignore this... Since in V2 it seems not to be an issue.

I'd make sure a patch for the 1.x would be clearly communicated in the PR description to avoid similar situation in the future 🙏🏻

arturmiglio (approved):

src/components/Pagination/Pagination.tsx
martimalek, line 138: This is a bug that was [already fixed in v2 by using `minWidth="5em"`](https://github.com/freenowtech/wave/blob/main/src/components/Pagination/Pagination.tsx#L145), would the same fix work for DMT or do you need to remove the `width` altogether?

JanHamara: okay 👍 the pagination is not optimised for mobile at all, which is separate tech debt itself that should be resolved, for now we're just updating this for DMT (so for Wave v1).

JanHamara: Without having any context, is there particular reason you set `em` there, not `rem`?

JanHamara: But yeah, we can use this fix instead for now to not cause overlapping of selector with pagination buttons 👍

martimalek: I don't recall the reasoning behind using `em` tbh 😅 I see you are using `rem` so I think it's fine

martimalek (approved):

src/components/Pagination/Pagination.tsx
JanHamara: @arturmiglio I've commited fix for the layout as well, to not cause overlapping of elements (like in v2)

PR #464, opened 2024-08-27
feat: add exports for experimental components

martimalek:
## What

Adds exports for the `experimental` folder, so that when distributing the package it can be accessed directly.

### Media

Now the experimental components can be accessed like:
```typescript
import { Button, Chip } from '@freenow/wave/experimental'
``` 

## Why

To allow to directly access experimental features, without needing to know the internal file structure of the package.

## How

- Adds [exports](https://nodejs.org/api/packages.html#exports) field in package.json to define the default + the experimental exports
- Adds `typesVersions` field to package.json to override the `experimental` types (so they don't come from the main index.d.ts file)

## Extra

I'm opening this as a draft PR because the experimental components are not ready to be distributed yet, once we have a first experimental component we can merge this.

I've already tested this approach in several `rc` releases, if you want to try it out as well you can check [2.6.0-rc.8](https://www.npmjs.com/package/@freenow/wave/v/2.6.0-rc.8). I've rolled back the version in package.json so that it gets automatically updated by semantic-release once the PR is merged.

The linting issues that you see are due to comments in some experimental components, those should be solved as part of the PRs for each component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,29 @@
 {
     "name": "@freenow/wave",
-    "version": "2.5.0",
+    "version": "2.6.0-rc.8",
     "description": "React components of the Wave design system for your Front-End project",
     "main": "lib/cjs/index.js",
     "typings": "lib/types/index.d.ts",
     "module": "lib/esm/index.js",
+    "exports": {
+        ".": {
+            "import": "./lib/esm/index.js",
+            "require": "./lib/cjs/index.js",
+            "default": "./lib/esm/index.js"
+        },
+        "./experimental": {
+            "import": "./lib/esm/experimental/index.js",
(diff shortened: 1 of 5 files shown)


dornelasnelson (approved):

package.json
```diff
@@ -44,8 +63,7 @@
         "url": "https://github.com/freenowtech/wave.git"
     },
     "files": [
-        "lib/",
-        "codemods/"
```
arturmiglio, line 48: Why remove codemods?

arturmiglio (approved):

package.json
alatielle: Not sure why it was done, I'll put them back

PR #469, opened 2024-09-19
Merge experimental components

alatielle:


src/components/experimental/Chip/Chip.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+function Chip({ label, isActive = false, Icon = PlusIcon, deletable = false }: ChipProps): ReactElement {
+    return (
+        <Container isActive={isActive}>
+            <>
+                {Icon && <Icon size={20} />}
+                <Spacer />
+                <Text variant="label1">{label}</Text>
+                <Spacer />
+                {deletable && (
```
dornelasnelson, line 77: The chips can also have another icon instead of the deletable one in the right, should we add it here or in a next ticket? Examples (calendar and FREENOW icons in the list)

dornelasnelson (approved):

src/components/experimental/Chip/Chip.tsx
alatielle: There's a ticket for Chips, this one is an initial WIP

PR #484, opened 2024-10-22
feat: adding an experimental divider component

lloydaf:
## What

This PR adds a divider component to the experimental components list

### Media

Screenshot 2024-10-22 at 17 36 29

​

## Why

DISP-144
​

## How

* Reuse the same API as the existing `Divider`
* Add a new experimental semantic token `divider` for the colours

### DIFF
--- a/src/components/experimental/Divider/Divider.spec.tsx
+++ b/src/components/experimental/Divider/Divider.spec.tsx
@@ -1,0 +1,35 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Divider } from './Divider';
+
+describe('Experimental: Divider', () => {
+    it('renders a horizontal divider by default when not passing any props', () => {
+        render(<Divider />);
+        expect(screen.getByTestId('horizontal-divider-experimental')).toBeInTheDocument();
+        expect(screen.queryByTestId('vertical-divider-experimental')).not.toBeInTheDocument();
+    });
+
+    it('renders a vertical divider when passing vertical prop', () => {
+        render(<Divider vertical />);
+        expect(screen.getByTestId('vertical-divider-experimental')).toBeInTheDocument();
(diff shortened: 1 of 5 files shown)


src/components/experimental/Divider/docs/Divider.stories.tsx
```diff
@@ -0,0 +1,131 @@
...
+            >
+                <Text>One</Text>
+                <Story />
+                <Text>Two</Text>
+            </div>
+        )
+    ]
+};
+
+export const VerticalInset: Story = {
```
alatielle, line 87: The weird thing is I couldn't find any vertical dividers with these variants in the wild :( I'd say let's merge it and see how it evolves?

alatielle (approved):

src/components/experimental/Divider/Divider.spec.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Divider } from './Divider';
+
+describe('Experimental: Divider', () => {
```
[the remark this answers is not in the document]
lloydaf: ❤️ 

We should have atleast some basic tests for every component imo

src/components/experimental/Divider/docs/Divider.stories.tsx
lloydaf: You know my feelings on the topic 😹 I honestly think that the variant in its entirety is pointless here (considering we can use space and layout props), so I don't mind either way.. 

but in this specific case, it's probably harder to enforce the idea that the "variant" only makes sense for horizontal dividers and that it doesn't do anything for vertical dividers (unless we treat them as an error perhaps), so I would suggest leaving this in, if we never turn out to use it, we can probably remove it later when we do a learned refactor

PR #487, opened 2024-10-31
fix: changing the cursor pointer for disabled states

lloydaf:
## What

I noticed that we are showing the cursor for disabled buttons. This MR fixes it

## Why

It can be confusing to the user to show a cursor when the button is not clickable
​

## How

* CSS

### DIFF
--- a/src/components/experimental/Button/Button.tsx
+++ b/src/components/experimental/Button/Button.tsx
@@ -62,6 +62,10 @@ const ButtonStyled = styled(BaseButton)<{ $emphasis: Emphasis }>`
 
     cursor: pointer;
 
+    &[data-disabled] {
+        cursor: not-allowed;
+    }
+
     &::before {
         position: absolute;
         top: 0;


src/components/experimental/Button/Button.tsx
```diff
@@ -62,6 +62,10 @@ const ButtonStyled = styled(BaseButton)<{ $emphasis: Emphasis }>`
 
     cursor: pointer;
 
+    &[data-disabled] {
+        cursor: not-allowed;
```
alatielle, line 66: I'd do "default" cursor, not-allowed is too dramatic :D

alatielle:
Nice catch!

src/components/experimental/Button/Button.tsx
lloydaf: Hehe I don't have a strong opinion on the subject, but since the [best practices](https://webreference.com/css/properties/cursor) suggests
> Use the "not-allowed" cursor for disabled elements to indicate that they are not clickable.

I thought it's better to keep it this way, considering we don't have any design specs for how it should look like. With the default pointer it kinda looks to me like the disabled element isn't there and that's not the best user experience imho

dornelasnelson (approved):

src/components/experimental/Button/Button.tsx
lloydaf: I will keep it for now @alatielle and mark this as resolved, there's a chance the designers agree with you, promise to patch it super quick in that case 🫰

PR #490, opened 2024-11-13
feat(dialog): add experimental dialog

alatielle:
## What

Adds an experimental Dialog Component using the new Modal & Backdrop components. I decided to not re-export them as public yet

### Media

​
image

## Why

​https://jira.intapps.it/browse/DISP-331

### DIFF
--- a/src/components/experimental/Dialog/Dialog.tsx
+++ b/src/components/experimental/Dialog/Dialog.tsx
@@ -1,0 +1,69 @@
+import React, { ReactElement, ReactNode } from 'react';
+import { Heading } from 'react-aria-components';
+import styled from 'styled-components';
+import { Text, textStyles } from '../Text/Text';
+import { Modal } from '../Modal/Modal';
+import { Backdrop, BackdropProps } from '../Backdrop/Backdrop';
+import { getSemanticValue } from '../../../essentials/experimental';
+
+const Card = styled.div`
+    display: grid;
+    gap: 0.5rem;
+`;
+
+const ButtonsWrapper = styled.div`
+    padding-top: 1.5rem;
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+    gap: 1rem;
+`;
+
+const StyledModal = styled(Modal)`
+    width: 30rem;
+`;
+
+const HeadlineText = styled(Heading)`
+    margin: 0;
+    ${textStyles.variants.headline}
+`;
+
+const SubtitleText = styled(Text)`
+    color: ${getSemanticValue('on-surface-variant')};
+`;
+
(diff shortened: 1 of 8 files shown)


src/components/experimental/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,69 @@
...
+const ButtonsWrapper = styled.div`
+    padding-top: 1.5rem;
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+    gap: 1rem;
+`;
+
+const StyledModal = styled(Modal)`
+    width: 30rem;
```
lloydaf, line 23: shouldn't this be controllable?

alatielle: In the dialog -- no, in a modal -- absolutely!

src/components/experimental/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface DialogProps extends Omit<BackdropProps, 'isDismissable' | 'isKeyboardDismissDisabled'> {
+    role?: 'dialog' | 'alertdialog';
+    headline: ReactNode;
+    subtitle: ReactNode;
+    dismissButton: ReactNode;
+    actionButton: ReactNode;
+}
+
+const Dialog = ({
+    role = 'dialog',
```
lloydaf, line 44: Should we allow other roles?

src/components/experimental/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,69 @@
...
+    dismissButton,
+    actionButton,
+    ...props
+}: DialogProps): ReactElement => (
+    <Backdrop {...props} isDismissable={false} isKeyboardDismissDisabled>
+        <StyledModal role={role}>
+            <Card>
+                <HeadlineText slot="title">{headline}</HeadlineText>
+
+                <SubtitleText as="p" variant="body1">
```
lloydaf, line 56: Maybe not within the scope of the PR, but can we name the variants with a more semantic name? `body1`, `body2` etc. don't help me understand what the variant does

src/components/experimental/Dialog/Dialog.tsx
alatielle: We do, It's either dialog or alertdialog

src/components/experimental/Dialog/Dialog.tsx
alatielle: It's named like the designers name it and what they use in the designs

dornelasnelson (approved):

PR #492, opened 2024-11-18
feat(select): add value customisation

alatielle:
## What

Adds customisation ability for the Select value

### Media

Screenshot 2024-11-18 at 18 10 12

## Why

Normally, select works like WYSIWYG: the user selects an item in the list and it is displayed as a select value

​Scenario: we might show additional information in the list (Fleet label + fare + ETA), but still want to display only the label as the value, when the item is selected

## How

Introducing the `renderValue` prop

### DIFF
--- a/src/components/experimental/Select/Select.tsx
+++ b/src/components/experimental/Select/Select.tsx
@@ -1,10 +1,11 @@
-import React, { ReactElement, useState } from 'react';
+import React from 'react';
 import {
     Select as BaseSelect,
     SelectProps as BaseSelectProps,
     SelectValue,
     SelectStateContext,
-    FieldError
+    FieldError,
+    SelectValueRenderProps
 } from 'react-aria-components';
 import { useIsSSR } from 'react-aria';
 import { useResizeObserver } from '@react-aria/utils';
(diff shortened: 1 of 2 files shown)


lloydaf (approved):

lloydaf:
I feel like this should be a configuration; i.e; we should have variants where this is possible and this is not possible (and would result in a validation error) depending on a config/prop value

But for the sake of the dispatcher I am fine with merging, but this is in "wave" so we should probably have this option in the future imo

lloydaf:
> > I feel like this should be a configuration; i.e; we should have variants where this is possible and this is not possible (and would result in a validation error) depending on a config/prop value
> 
> What scenario do you have in mind?

A normal scenario for a typeable select box component where we're only supposed to use the values available to us; that's how normal selects work, no?

alatielle:
> A normal scenario for a typeable select box component where we're only supposed to use the values available to us; that's how normal selects work, no?

It still uses the values available in the items, this is just a presentation change for the selected item

lloydaf:
> > A normal scenario for a typeable select box component where we're only supposed to use the values available to us; that's how normal selects work, no?
> 
> It still uses the values available in the items, this is just a presentation change for the selected item

Respectfully I don't see the point in this change then; what are we trying to achieve? Can you add a more "real-life" example to the docs to illustrate where this is useful better?

alatielle:
> Can you add a more "real-life" example to the docs to illustrate where this is useful better?
Absolutely!
Imagine this: we have an additional information in the list to help make a selection
Screenshot 2024-11-19 at 14 41 40
But we want to display only the label as the select value, so in this case "Taxi XL", not "Taxi XL in 6 min 15-17 EUR"
Screenshot 2024-11-19 at 14 41 54

This is what this change is for

dornelasnelson (approved):

PR #504, opened 2024-12-04
fix: preserve button content to maintain consistent width during loading state

AlanFnz:
## What

Preserves button width during loading state by wrapping children and spinner in separate containers, hiding children with opacity and visibility

### Media

https://github.com/user-attachments/assets/11f3e29a-f4d2-49ff-b3f9-651cacea467c

https://github.com/user-attachments/assets/1c3f1562-bca3-41c4-86d0-7bd832c5e3ff

## Why

Improves the layout consistency


alatielle (approved):
:shipit:

PR #506, opened 2024-12-27
feat(combo-box): external input ref support

AlanFnz:
## What

This PR adds support for an external `ref` in the `ComboBox` component, enabling parent components to pass their own refs for access to the input element within the `ComboBoxInput` 

### Media

https://github.com/user-attachments/assets/18122ae2-ffbf-48ca-b317-5dbd1fa4db18

## Why

So we can programmatically focus or interact with the input field within the `ComboBox`
​
A requirement of UI improvement SUP-45087

### DIFF
--- a/src/components/experimental/ComboBox/docs/ComboBox.stories.tsx
+++ b/src/components/experimental/ComboBox/docs/ComboBox.stories.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import { StoryObj, Meta } from '@storybook/react';
 import { ComboBox } from '../ComboBox';
 import { ListBoxItem } from '../../ListBox/ListBox';
+import { Button } from '../../../Button/Button';
 import DogIcon from '../../../../icons/basic/DogIcon';
 
 const meta: Meta = {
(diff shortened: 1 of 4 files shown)


src/utils/hooks/useMergeRefs.ts
```diff
@@ -0,0 +1,21 @@
+export default function useMergeRefs<T>(
+    ...inputRefs: (React.Ref<T> | undefined)[]
+): React.Ref<T> | React.RefCallback<T> {
+    const filteredInputRefs = inputRefs.filter(Boolean);
```
lloydaf, line 4: Is this the same as `inputRefs.filter(ref => ref)`?

src/components/experimental/ComboBox/docs/ComboBox.stories.tsx
```diff
@@ -153,3 +154,81 @@ export const FullyControlled: Story = {
...
+            return () => {
+                ignore = true;
+            };
+        }, [filterText]);
+
+        const handleFocus = () => {
+            if (comboBoxRef.current) {
+                comboBoxRef.current.focus();
+                // eslint-disable-next-line no-console
+                console.log('input focused');
```
lloydaf, line 188: do we need these logs?

lloydaf (approved):

src/utils/hooks/useMergeRefs.ts
AlanFnz: Yes, its functionally the same, Boolean would be in this case kind of a shorthand for filtering out falsy values. I can update it to use `ref => ref` for explicitness if that's preferred. Let me know!

src/components/experimental/ComboBox/docs/ComboBox.stories.tsx
AlanFnz: Hmm not really, in the case of focus and blur, the change in the input state is evident. As for `handleGetValue` maybe we can remove it completely, the focus and blur actions are demonstration enough probably.

src/utils/hooks/useMergeRefs.ts
lloydaf: I think it's fine the way it is.. thanks!

src/components/experimental/ComboBox/docs/ComboBox.stories.tsx
lloydaf: Oh this is for stories.. nevermind

PR #507, opened 2025-01-08
Fix: vertical divider height

dornelasnelson:
## What

Fixes the Vertical Wave Divider to not have a height of 100% and stretch instead, so that each of the variants have their correct margins and work as expected

### Media

image


alatielle:
`yarn.lock` shouldn't be a part of this MR, check pls

src/components/experimental/Divider/Divider.tsx
```diff
@@ -53,8 +53,7 @@ const HorizontalLine = styled.div<DividerProps>`
 `;
 
 const VerticalLine = styled.div<DividerProps>`
-    display: inline-block;
-    height: 100%;
+    align-self: stretch;
```
alatielle, line 56: What if the parent is not a flex/grid element?

dornelasnelson: 🤔 
I honestly don't know. Tried changing one of the storybooks to display block and brining back the inline-block here but also doesn't work

PR #508, opened 2025-01-08
fix: divider margin

dornelasnelson:
## What

Fixes Divider margins to 0.5rem instead of 1rem as it was too big.

### Media

The only place we are currently using looked like this with 1rem:
image

With 0.5rem it looks as expected:
image


lloydaf (approved):
Can we override this programmatically? I feel like maybe we should make this more flexible

dornelasnelson:
> Can we override this programmatically? I feel like maybe we should make this more flexible

yes @lloydaf , we can override it! it's pretty flexible but I thought it would be better if it was with the 0.5rem as the first usage of it would already be overwritten

PR #512, opened 2025-01-24
feat(table): add pointer as a cursor for interactive rows

alatielle:
## Why

​To better indicate that the row is interactive

### DIFF
--- a/src/components/experimental/Table/Table.tsx
+++ b/src/components/experimental/Table/Table.tsx
@@ -81,6 +81,7 @@ const Row = styled(BaseRow)`
     &[data-hovered] td::before {
         background: ${getSemanticValue('on-surface')};
         opacity: 0.08;
+        cursor: pointer;
     }
 
     &[data-selected] {


src/components/experimental/Table/Table.tsx
```diff
@@ -81,6 +81,7 @@ const Row = styled(BaseRow)`
     &[data-hovered] td::before {
         background: ${getSemanticValue('on-surface')};
         opacity: 0.08;
+        cursor: pointer;
```
dornelasnelson, line 84: thinking on a table that the rows are not "selectable" does it make sense that it has the pointer cursor? idk if we can check here that it has the "onAction" function set in order to set the cursor

alatielle: it won't have the `data-hovered` state if its not marked as `selectionMode!=="none"`

dornelasnelson: perfect 😄

PR #514, opened 2025-01-31
fix: add label prop to icon button

dornelasnelson:
## What

Adds a required label prop to <IconButton /> that adds a hidden text to the icon button to describe the Icon action.


shiba-codes (approved):
:shipit:

PR #515, opened 2025-02-04
fix: improve spaces when label is not defined (#450)

### LINKED ISSUE #450: Remove Checkbox margin when label is not defined
-   `@freenow/wave` version: 1.34.4

### What

Remove Checkbox margin when label is not defined, so that it is correctly aligned when used in UI with other form elements (but  without labels)

Screenshot 2024-05-23 at 11 53 56

### Reproduction

You can check this issue on registration drafts form on DMT

kabarchonok:
## What
Replaced `margin` with `gap` in Checkbox, RadioButton and Toggle components to improve UI when no label is undefined.

Also slightly changed the default history for the Checkbox similar to other components, made the default label undefined.

## Why

Closes #450

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -38,7 +38,7 @@ const WithTapAreaWrapper = styled.div<Pick<CheckboxProps, 'textVerticalAlign'>>`
     align-items: center;
 
     height: 1rem;
-    margin: 0 0.5rem 0 0;
+    margin: 0 0 0 0;
     ${({ textVerticalAlign }) => (textVerticalAlign === 'top' ? 'margin-top: 0.1875rem' : undefined)}
 `;
 
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -38,7 +38,7 @@ const WithTapAreaWrapper = styled.div<Pick<CheckboxProps, 'textVerticalAlign'>>`
     align-items: center;
 
     height: 1rem;
-    margin: 0 0 0 0;
+    margin: 0;
     ${({ textVerticalAlign }) => (textVerticalAlign === 'top' ? 'margin-top: 0.1875rem' : undefined)}
 `;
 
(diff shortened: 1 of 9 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -38,7 +38,7 @@ const WithTapAreaWrapper = styled.div<Pick<CheckboxProps, 'textVerticalAlign'>>`
     align-items: center;
 
     height: 1rem;
-    margin: 0 0.5rem 0 0;
+    margin: 0 0 0 0;
```
dornelasnelson, line 41:
```suggestion
    margin: 0;
```

src/components/RadioButton/components/Checkmark.tsx
```diff
@@ -15,7 +15,7 @@ const Checkmark = styled.input<CheckmarkProps>`
     width: 1rem;
     height: 1rem;
     padding: 0;
-    margin: 0 0.5rem 0 0;
+    margin: 0 0 0 0;
```
dornelasnelson, line 18:
```suggestion
    margin: 0;
```

src/components/Toggle/Toggle.tsx
```diff
@@ -22,6 +23,11 @@ interface ToggleProps extends ToggleHtmlInputProps, ClassNameProps, MarginProps
     error?: boolean;
 }
 
+const SlideWrapper = styled.span`
```
dornelasnelson, line 26: Is this wrapper added so that the gap on the label would be between this wrapper and the actual label?

dornelasnelson (commented):
Hey @kabarchonok, thanks a lot for contributing to wave. I've left a few comments and  suggestions. I'm talking with the team to understand whether this is really relevant as the issue was about wave version 1 and we are at version 2 now but I ran your forked repo here and it looks like it works well.

src/components/Toggle/Toggle.tsx
kabarchonok: That's right, the Wrapper was added to avoid the extra gap that is caused by the invisible input, since the flex parent has three children. 

In my opinion the `gap` approach would be more simple declarative than adding some condition to set `margin`

kabarchonok:
@dornelasnelson You're welcome! I just did it for fun, so I don't pretend to be opinionated.

dornelasnelson (approved):

AlanFnz (approved):

PR #520, opened 2025-02-24
fix: lint warnings

AlanFnz:
## What

This PR addresses and resolves the current warnings we are getting when running our linter

## Why

Technical debt
​
## How

It consists of:
- Adding missing type annotations, mostly in stories
- Removing unused imports
- Removing alert from the experimental `IconButton` story in order to comply with the `no-alert` rule
- Fix the `placement` property used in the `Tooltip` story
- Use media query list obtained with `evalColorSchemeQuery` instead of the event's list in `getCurrentScheme`
  - This ensures the correct type (MediaQueryList) is passed and reflects the current state
- Replace deprecated rules with their current equivalents in stylelint config

### DIFF
--- a/src/components/ColorScheme/docs/CurrentScheme.tsx
+++ b/src/components/ColorScheme/docs/CurrentScheme.tsx
@@ -5,13 +5,13 @@ import { Text } from '../../Text/Text';
 const evalColorSchemeQuery = () => window.matchMedia('(prefers-color-scheme: dark)');
 const getCurrentScheme = (query: MediaQueryList): 'dark' | 'light' => (query.matches ? 'dark' : 'light');
 
-export const CurrentScheme = () => {
+export const CurrentScheme: React.FC = () => {
     const [currentScheme, setCurrentScheme] = useState(getCurrentScheme(evalColorSchemeQuery()));
 
     useEffect(() => {
         const mql = evalColorSchemeQuery();
-        const onMediaQueryChange = (e: MediaQueryListEvent) => {
-            setCurrentScheme(getCurrentScheme(e));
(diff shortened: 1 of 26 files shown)


src/components/ColorScheme/docs/CurrentScheme.tsx
```diff
@@ -5,13 +5,13 @@ import { Text } from '../../Text/Text';
 const evalColorSchemeQuery = () => window.matchMedia('(prefers-color-scheme: dark)');
 const getCurrentScheme = (query: MediaQueryList): 'dark' | 'light' => (query.matches ? 'dark' : 'light');
 
-export const CurrentScheme = () => {
+export const CurrentScheme: React.FC = () => {
     const [currentScheme, setCurrentScheme] = useState(getCurrentScheme(evalColorSchemeQuery()));
 
     useEffect(() => {
         const mql = evalColorSchemeQuery();
-        const onMediaQueryChange = (e: MediaQueryListEvent) => {
```
AlanFnz, line 13: As mentioned in the PR description, this change ensures the correct type is passed to `getCurrentScheme` and also reflects the current state of the media query.
Besides that, I couldn't find any usage of this component. I don't think storybook is somehow pickup this up, but Im not 100% sure. If that's the case maybe we can just remove `CurrentScheme` entirely.

shiba-codes (approved):
Amazing 🚀

src/components/ColorScheme/docs/CurrentScheme.tsx
AlanFnz: Created [this issue](https://github.com/freenowtech/wave/issues/521) to revisit this

PR #522, opened 2025-03-05
feat: add search component

shiba-codes:
## What

This is an experimental Search component. Additionally, I changed the behavior of the clear button in TextField to be the same of the Search

### Media

Capture-2025-03-05-175811
Capture-2025-03-05-175839
Capture-2025-03-05-175908​

## Why

​


renejfc:
LGTM! 👍

PR #525, opened 2025-04-23
feat: experimental checkbox component

AlanFnz:
## What

Based on the existent Checkbox component, adds the experimental version.

### Media

indeterminate

invalid-selected

disabled-indeterminate

disabled-checked

disabled

error

selected

default

## Why

Required for the GTC flow implementation SUP-47661
​


src/components/experimental/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,185 @@
...
+        background-color: ${getSemanticValue('negative')};
+        border-color: ${getSemanticValue('negative')};
+    }
+`;
+
+const Checkbox: FC<CheckboxProps> = props => {
+    const { isDisabled, isInvalid, isIndeterminate, label, variant = 'body1', ...rest } = props;
+
+    let dynamicLabel: ReactNode = label;
+    if (typeof label === 'string') {
```
dornelasnelson, line 159: Why does the label have a ReactNode type instead of a string? Is there any particular usage that it would receive a node instead? In all tests / stories it shows as string as well, so I'm thinking if it could just be changed the type to optional string on line 16 and then not have this if here 🤔

dornelasnelson (approved):

src/components/experimental/Checkbox/Checkbox.tsx
AlanFnz: The idea is to support a React Node so we can use more complex label content, like links, rich text formatting, icons or other components. The non-experimental version of the Checkbox also follows this approach.

dornelasnelson: alright! sounds good 😃

PR #530, opened 2025-05-08
fix: avoid isLoading prop on button element

jonotrujillo:
## What

When using the experimental `Button` component, the following error is show in the browser terminal:

```
React does not recognize the `isLoading` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isloading` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

## Why

The `isLoading` prop is accidentally passed to the `button` element, which is not recognized.

## How

This issue is resolved by utilizing the transient props feature provided by styled-components (https://styled-components.com/docs/api#transient-props).


dornelasnelson (approved):
🤩

PR #531, opened 2025-05-08
feat: adding an experimental tooltip component based on react-aria

lloydaf:
## What

In this MR, we are adding an experimental version of a tooltip component

### Media

​
Screenshot 2025-05-08 at 14 58 31

Screenshot 2025-05-08 at 14 58 39

## Why

This is required for the dispatcher tool​

## How

* extend tooltip from `react-aria`
* add some basic custom behaviour

### DIFF
--- a/src/components/experimental/Tooltip/Tooltip.tsx
+++ b/src/components/experimental/Tooltip/Tooltip.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import { OverlayArrow, Tooltip as AriaTooltip, TooltipTrigger, Focusable } from 'react-aria-components';
+
+import type { TooltipProps as AriaTooltipProps, TooltipTriggerComponentProps } from 'react-aria-components';
+
+import './tooltip.css';
+
+interface TooltipProps {
+    children: any;
+    content: string;
+    hideArrow?: boolean;
+    customTrigger?: boolean;
+    tooltipProps?: Omit<AriaTooltipProps, 'children'>;
+    triggerProps?: Omit<TooltipTriggerComponentProps, 'children'>;
+}
+
+const Tooltip = ({
+    children,
+    content,
+    triggerProps,
+    tooltipProps,
+    hideArrow = false,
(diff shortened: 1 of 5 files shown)


src/components/experimental/Tooltip/Tooltip.tsx
AlanFnz, line 1: Would it make sense to add some tests for this component?

src/components/experimental/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { OverlayArrow, Tooltip as AriaTooltip, TooltipTrigger, Focusable } from 'react-aria-components';
+
+import type { TooltipProps as AriaTooltipProps, TooltipTriggerComponentProps } from 'react-aria-components';
+
+import './tooltip.css';
+
+interface TooltipProps {
+    children: any;
```
AlanFnz, line 9: nit: maybe `ReactNode` would be safer?

AlanFnz (approved):

src/components/experimental/Tooltip/Tooltip.tsx
lloydaf: Screenshot 2025-05-08 at 18 47 13
I tried that, I am getting this error message :(

src/components/experimental/Tooltip/Tooltip.tsx
lloydaf: It is a bit tricky to test tbh, `user.hover` does not seem to work properly to test for the tooltip

However, I have added a basic test (with the tooltip open by default) to add some minimum assurance, we can enhance the test suite later once we discover a reliable way to test this

src/components/experimental/Tooltip/Tooltip.tsx
lloydaf: I will mark this as resolved for the moment, for the sake of moving on

PR #554, opened 2025-08-06
fix: pass hideLabel prop to TimeField InnerWrapper

AlanFnz:
## What

Properly controls top padding in `TimeField` by passing `hideLabel={!label}` to `InnerWrapper`. Padding is now only applied when a label actually exists

### Media

timefield


renejfc (approved):
lgtm

PR #565, opened 2025-10-29
fix: changed size to match padding of other icons

lizzielerwill-ctrl:
## What

- Corrected the size of the edit icon so it had enough padding to match others


renejfc:
lgtm!

